PR #4, opened 2022-08-24
[WIP] Feature - Support multiple batches, change default prompt text to placeholder, clear images array before each new generation

EuroWhisper:


EuroWhisper:
Hmm... with regard to implementation of batches, the seed will potentially be different for each batch, but on the FE it will only list a single seed. I need to give this more thought.

EuroWhisper:
@cmdr2 Hi! I'm thinking with batches, maybe I can visually separate/group each batch somehow, with each batch group displaying its seed... What do you think?

cmdr2:
Hi @EuroWhisper thanks for the PR! From what I understand, this is letting you create multiple images in one go, using the same prompt. Is that different from the current `num_outputs` field?

The `num_outputs` field currently lets you do 1 or 4 images in one go, but I see that as an artificial restriction put by the docker container. I'm looking at ways to bypass that restriction, so we can generate more or less than just 1 or 4 images.

The disable sound is a good idea.

EuroWhisper:
@cmdr2 Good morning. Thanks for taking a look. Yes, it lets you create multiple images in one go, using the same prompt/settings. It will output a number of images corresponding to `num_batches * num_outputs`. So if for example you set `num_batches = 3 `and `num_outputs = 4` you'll end up with 12 images in total, with each batch containing 4 images.. and it will generate 1 batch at a time (4 images at a time in this case). 

My situation is I have a 3070 ti card and I can't generate 4 images at once (just 1). So when I find a good prompt, I'm setting `num_batches = 50` and `num_outputs = 1`, so I can get 50 images in the end, without having to press the "make image" button 50 times in a row. This is also why I added an option to disable sound, so that it won't ding 50 times in a row in my case. :)

cmdr2:
Thanks, that makes sense. I will think about this a bit more. It's getting a bit late where I am, so I will respond tomorrow :)

dsmmcken:
I also can only generate 1 at a time, so I like this idea. Might be nice if it just continually generates until you say to stop?

cmdr2:
Hi, I've added this feature in 6b55f385c73b88c34857d9f266c0c4f69df6a158

It allows you to set the total number of images, and then pick the "parallel count" (of 1 or 4) depending on the capabilities of your GPU. So if you can generate 4 at a time, then 8 total images will only take 2 requests to process.

The 'ding' completion sound will play only once all the tasks are complete.

Thanks @EuroWhisper and everyone for their suggestion  :)

PR #6, opened 2022-08-25
Feature - Add audio toggle with value that persists between page loads

EuroWhisper:


EuroWhisper:
@cmdr2 Here's the audio toggle PR as requested :)

index.html
```diff
@@ -68,10 +69,20 @@
 </body>
 
 <script>
+const SOUND_ENABLED_KEY = "soundEnabled"
+document.querySelector('#sound_toggle').checked = isSoundEnabled();
```
cmdr2, line 73: Can you please move this `.checked` setter to line 245, so the consts can stay at the top and actions at the bottom?

index.html
```diff
@@ -45,6 +45,7 @@
 
     <div id="configHeader"><b>Advanced settings:</b> [<a id="configToggleBtn" href="#">show</a>]</div>
     <div id="config">
+        <input id="sound_toggle" name="sound_toggle" type="checkbox" checked> <label for="sound_toggle">Enable Sound</label><br/>
```
cmdr2, line 48: Can you please add this as the last setting, instead of the first? Seed, num outputs, etc are used more frequently.

Also, how about naming this "Play sound on completing the task"? Probably clearer to people what the sound option is for. Thanks!

cmdr2:
Looks good! Thanks for this. Just added three minor comments, but looks good to go!

EuroWhisper:
@cmdr2 Thanks! Feel free to merge whenever you're ready :)

cmdr2:
Thanks!

PR #7, opened 2022-08-25
use HTTPS for repo clone instead

lumir-mrkva:
ssh won't work for everyone


cmdr2:
Good catch! Thanks, merged.

PR #19, opened 2022-08-27
Allow other image file types

dsmmcken:
The underlying library accepts a wide range of filetypes, no need to restrict it to just png. I wanted to use a jpg.


cmdr2:
Looks good, thanks!

PR #29, opened 2022-08-30
Add save pictures

heylenz:
When we generate images, we may want to save all the pictures we generate and keywords


cmdr2:
Hi @heylenz thanks for this! I will look into this in more detail tomorrow.

Quick question - would this save the images inside the docker container? Would we need to use a docker command to copy the image to the host computer? Thanks

GregHilston:
I was about to request this feature, and I'm glad I saw this PR.

it'd be lovely to simply check a box off, and have our web browser download the image(s) generated, with the prompts as the filename.

GregHilston:
@heylenz I forgot to comment on the source code of this PR itself.

I'm with @cmdr2, that we'd need to provide a volume from the host OS to the container, so the images persisted. Additionally, would you be open to instead of automatically downloading the file to the server, instead having it be a button that each image displayed, and was clickable? That way the user can pick and choose what to download. Additionally, it'd be nice to instead of randomly generating the file name, having the prompt be in the file name somehow. Thoughts?

I'm happy to submit these changes into the code itself, if you agree with them.

cmdr2:
@GregHilston There is a download button next to each image currently.

I've actually added this to v2, with the images being saved to the user's home folder ("C:\Users\fooUser\Stable Diffusion UI" on Windows, and "/home/fooUser/Stable Diffusion UI" on Linux), with the prompt name being used for the folder containing the images. New requests with the same prompt will go into the same folder, with increasing numbers being added to their filename.

There is a setting in the "Advanced Settings" UI to automatically save the images to disk, which will trigger this behavior.

I'm about to commit this change to v2 within a few hours, so I would suggest holding off on this PR for now. I've credited @heylenz  for this in the commit.

Apologies for going ahead and making the change unilaterally, the v2 codebase is evolving rapidly. Thanks!

PS: There's a discord server now! For development discussions and support: https://discord.com/invite/u9yhsFmEkB . Would love to see you there!

cmdr2:
@GregHilston @heylenz - I've updated v2 with this feature (and credited @heylenz in the commit message). https://github.com/cmdr2/stable-diffusion-ui/commit/74d04f2cdb52074fbfed8dce93289a6d05dacea1

It works as described above. An option to save the images to the user's home folder (disabled by default). It saves by using the prompt for the folder name. And each image can be saved individually by clicking the "Download" button overlaid on the image in the browser.

You can download it from https://drive.google.com/file/d/1cEuOcb9OaldXcc2XzEMqEdvAr_w-KJ_p/view?usp=sharing and then double-click Start Stable Diffusion UI.cmd to install and run.

Please let me know if there are any problems after running it.

PS: There's a new Discord server for support and development discussions: https://discord.com/invite/u9yhsFmEkB . Please join in for faster discussion and feedback on v2. Thanks!

heylenz:
> Hi @heylenz thanks for this! I will look into this in more detail tomorrow.
> 
> Quick question - would this save the images inside the docker container? Would we need to use a docker command to copy the image to the host computer? Thanks


We don't need to use additional docker commands,
In docker-compose.yaml we have a volume map.
    volumes:
      - .:/app
      -
in main.py, I save the pictures to OUTPUT_DIR 
OUTPUT_DIR = "/app/output"

So we can see that the image is directly generated
image

I originally had an idea to make a separate page to display all our generated images on the web page. When the mouse is placed, we can see the keywords we used to generate this image.


Each picture corresponds to a metadata, which is convenient to go back to the history,
`
{
  "input": {
    "prompt": "......,Cartoon, Comic Book, Concept Art, Digital Art, by Andy Warhol",
    "num_outputs": "1",
    "num_inference_steps": "50",
    "width": "512",
    "height": "512",
    "seed": "6193",
    "guidance_scale": "7.5"
  }
}`

heylenz:
> @heylenz I forgot to comment on the source code of this PR itself.
> 
> I'm with @cmdr2, that we'd need to provide a volume from the host OS to the container, so the images persisted. Additionally, would you be open to instead of automatically downloading the file to the server, instead having it be a button that each image displayed, and was clickable? That way the user can pick and choose what to download. Additionally, it'd be nice to instead of randomly generating the file name, having the prompt be in the file name somehow. Thoughts?
> 
> I'm happy to submit these changes as a branch, if you agree with them.

Answers to random names,
I was thinking that we might continue to generate a lot of images, some are just individual keywords, some are image2image, and the keywords can be very long, so my idea is to use a page to display, and then have a picture detail, you can Accurately know the keywords of each image, as well as the generation process, including image generation.

Please feel free to change,

cmdr2:
Hi @heylenz Thank you for the clarification! I'm sorry I didn't respond earlier, I was engaged fully with trying to get v2 released.

As you know v2 doesn't use docker anymore, so unfortunately this particular PR is no longer applicable. I will close it in favor of this ticket for a "Gallery / Image Browser" (the idea you described above). I really like the idea of a gallery, and was wondering about it just today morning. https://github.com/cmdr2/stable-diffusion-ui/issues/56

If you're interested, I'd love to accept a PR for a quick and simple image gallery. The new `develop` branch is the place to develop new features, and I'm going to add instructions there for how to run and test your changes locally (during development).

Again, apologies for closing this PR, but essentially the ideas related to saving images have been added (credited to you in the commit log).

Thanks again!

GregHilston:
@cmdr2 may I ask why docker is no longer used?

I'm specifically curious because the rpeistiory having a docker compose file is what convinced me to try it out in the first place. Due to how quick and easy setting up a container is.

cmdr2:
Hi @GregHilston , v2 is a 1-click installer, by automatically setting up a conda environment for you, downloading all the necessary files and starting the interface. Interestingly, you don't even need to have conda installed previously, the installer has a packaged version of conda inside it. It doesn't install conda, so you'll never even need to know it exists.

So this enables anyone (with no technical knowledge) to download and use Stable Diffusion. That doesn't mean it is dumbed down for power users, because people can edit and replace the downloaded Stable Diffusion codebase (git cloned locally from the original github repo).

And the conda environment downloads the packages to a local folder (using the --prefix option), so it doesn't interfere with any other packages that you may have installed.

For Windows users, this removes the need for WSL and Windows 11, and the need to install the nvidia-container-toolkit package. It now works natively on Windows 10. WSL was required because nvidia-container-toolkit requires WSL.

All in all, IMO this achieves a decent degree of isolation (due to the --prefix local installation), while surpassing the docker approach in terms of ease of installation and dependencies. Some users installing v1 (docker) asked me what a terminal is. There's no reason why they can't use Stable Diffusion. Knowing what a terminal is doesn't need to be a gating function.

I understand the isolation guarantee of Docker, but this made it much easier for people to get started (based on what users have told me). :)

Thanks!

GregHilston:
Totally understood. Personally, I avoid Conda on my host OS at all costs, which is why the Docker isolation was nice, but I get what you're saying, and can see how this lowers the bar of entry for those less technical. 

Thanks for all you do!

cmdr2:
Thanks! :) As such you don't need to know this installs conda. The docker container in v1 also used conda to bootstrap the project, and this is similar, in that you don't need to know it uses conda.

PR #40, opened 2022-09-03
Additional modifiers

MrManny:
This is a small set of additional modifiers that have been added. In addition, modifiers were sorted alphabetically within their category (although I did it by hand and I may have missed something).


cmdr2:
Thanks a lot! :)

PR #44, opened 2022-09-03
updated filename

caranicas:
Fixing the issue I just created https://github.com/cmdr2/stable-diffusion-ui/issues/43. I love this project and just thought this would be a nice QOL improvement. 

If you would like to see a more individualized approach the the metadata adding with some additional checkbox / toggle UI, I am sure I could add that next.

PR #45, opened 2022-09-03
updated filenames for v2

caranicas:
A sibling PR to https://github.com/cmdr2/stable-diffusion-ui/pull/44. When I forked I wasn't really thinking about it, so my first branch was off main, then when I made the PR I realized I wanted to update V2, since that is what I was using, and I think where you probably will continue development. 

Wasn't sure what your merge strategy was so I figured I would make another PR to make it easier to eventually merge down.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -551,7 +551,7 @@ async function doMakeImage(reqBody) {
 
         imgSaveBtn.addEventListener('click', function() {
             let imgDownload = document.createElement('a')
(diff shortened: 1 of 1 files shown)


cmdr2:
Looks great! Thanks for this PR! :)

cmdr2:
Actually, : (color) isn't an allowed character on windows. Replacing that with -

PR #47, opened 2022-09-03
fixed support for mutli images

caranicas:
I didn't expect you to merge that PR so quickly, as I was doing some more testing. I realized I missed an edge case. I was pulled the seed value from the input and that would be wrong if you generated 2+ images. 

This new implementation adjust the logic to get the seed value from the label that is sibling to the download button. 

I also adjusted a bit of the formatting, I added the dimensions, which you could get by inspecting, but having it in the file name seemed nicer. I also  replaced the underscores with commas since I felt like that made more sense. 

Took a quick look to see if this was going to mess up the auto download but that still works fine.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -549,9 +549,9 @@ async function doMakeImage(reqBody) {
             seedField.disabled = false
         })
 
-        imgSaveBtn.addEventListener('click', function() {
+        imgSaveBtn.addEventListener('click', function(e) {
             let imgDownload = document.createElement('a')
-            imgDownload.download = createFileName();
+            imgDownload.download = createFileName(e);
             imgDownload.href = imgBody
             imgDownload.click()
         })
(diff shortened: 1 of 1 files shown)


caranicas:
Also realizing that windows is unhappy with `|` and `:` as part of file names, so Ill give that some thought.

cmdr2:
Hi @caranicas yeah I fixed the special character before releasing v2 to the main branch. Please check the `main` or `develop` branch's `index.html` for the change. Yes, you're right, the seed picked up is wrong though, hadn't noticed that.

Sorry for the inconvenience, but the `v2` branch has been deprecated after the public release, and the development will be on the `develop` branch now. Once changes on the `develop` branch look stable, they'll get merged into the `main` branch (which is used for auto-update).

If it's not too much trouble, could you please make your changes on the `develop` branch? Thanks

caranicas:
I'm heads down on the app structure work, but ill try to get to this, if no one else does.

caranicas:
replaced by

PR #54, opened 2022-09-05
v2.06

cmdr2:


Karric:
In addition to the warning, what about a button to downsample the image to something safe? Would save the user from having to do that manually using some other program.

PR #59, opened 2022-09-05
Download buttons

cmdr2:


cmdr2:
Also changed the download URLs to github

PR #72, opened 2022-09-06
More modifiers

SaulDoesCode:


ui/modifiers.json
```diff
@@ -2,7 +2,7 @@
   [
     "Drawing Style",
     [
-      "Cel Shading",
+      "Cell Shading",
```
cmdr2, line 5: Hi, it is actually Cel Shading :) https://en.wikipedia.org/wiki/Cel_shading

cmdr2:
Hi @SaulDoesCode Thanks for the PR! Nice work.

Question - if you haven't already done this, can you please test that these modifiers actually change the image to the desired style?

It would be good to avoid adding modifier tags that don't have an effect. Thanks!

ui/modifiers.json
```diff
@@ -127,24 +153,38 @@
...
       "by Caravaggio Michelangelo Merisi",
+      "by Claude Monet",
       "by David Mann",
+      "by Diego Velázquez",
+      "by Édouard Manet",
       "by Frida Kahlo",
+      "by Georgia O'Keeffe",
       "by H.R. Giger",
       "by Hayao Miyazaki",
+      "by Henri Matisse",
```
SaulDoesCode, line 164: Mattise style works I tried it

SaulDoesCode:
I did some renders on my machine with most of the new modifiers, they seem to yield agreeable results that match the meaning, I posted a few renders on the art section of the discord. I think most of the modifiers will have some effect, I have tried some fairly obscure artists and got good results, I think it should be fine. It's up to you.

cmdr2:
Thanks @SaulDoesCode , I've merged this in. Thanks for the PR! :)

PR #80, opened 2022-09-06
React bare metal

caranicas:
This should be safe to merge, and keep any conflicts or regressions to a minimum moving forward.


caranicas:
replaced by https://github.com/cmdr2/stable-diffusion-ui/pull/191

PR #83, opened 2022-09-07
generating in parallel not updating seed correctly.

ArtificialLegacy:


MrManny:
To elaborate on the underlying issue that hasn't been described in the initial PR: if you generate a number of images with a batch size of more than one and smaller than the number of total images to generate, the seed is calculated incorrectly.

For example, if you generate 6 images with a batch size of 3, the seeds would be [x, x+1, x+2], [x+1, x+2, x+3] as the seed for any subsequent batch will always start at +1 of the previous batch. This PR fixes it by incrementing the seed by the batch size.

MrManny:
image

Case in point, the first row and the second row each form their own batch. As you can see, there are duplicates between the batches.

cmdr2:
Thank you @ArtificialLegacy and @MrManny for this! Nice catch!

PR #105, opened 2022-09-07
Update modifiers.json

frode:
Added missing artists to the list of available styles


cmdr2:
Thanks @frode :) Quick question - have you tested that each of these modifiers produce the desired effect on the image?

Thanks

frode:
> Thanks @frode :) Quick question - have you tested that each of the new modifiers produce the desired effect on the image?
> 
> Thanks

Welcome :) Yes I tested every one of them before being confident to add them to the list.

I used this site as a reference for testing and validation of the expected art style:
https://weirdwonderfulai.art/resources/stable-diffusion-micro-artist-studies/

Some had to be verified by searching the artist name and looking at search results, example Jack Kirby.

While adding them, I figured it would have been nice to include one image per style to show in the web UI somehow, so that you can have a quick peek at the expected styles.

/ frode

frode:
By going through this page: https://weirdwonderfulai.art/resources/disco-diffusion-70-plus-artist-studies/
We could probably verify and add quite a few more artist styles to the available modifiers, but I do suspect that the way the UI is designed in its current state could need some adjustments before lots of more options gets added. 

What do you think? (I might be thinking premature optimization)

cmdr2:
Merged, thanks! :)

PR #139, opened 2022-09-10
fix: change to the correct working directory

mnahkies:
changes to the directory containing `start.sh` prior to activating the conda environment

this allows you to run the program without first changing to the correct directory, eg: `$ ~/bin/stable-diffusion-ui/start.sh`


cmdr2:
Thanks! :) Apologies for the long delay in merging, this is helpful.

PR #141, opened 2022-09-10
Update CONTRIBUTING.md to instruct commenting out multiple lines

youmslinky:
Need to comment out most of the lines so the `ui` folder symlink doesn't get removed and the `.sh` and `.bat` file changes don't get rolled back on next startup.


cmdr2:
Thanks! :)

PR #142, opened 2022-09-10
Update modifiers.json

thomasbaijot:
Added more artists and some other items, created "Mangakas" section and "Art Movements" section.
Sections are a bit messed up and overlap sometimes, i suggest to redefine them in a more logical way.
This is my first use of GitHub so be nice please :D


cmdr2:
Thanks! That's a big addition! Quick question - have you confirmed that each modifier creates the desired effect?

Just trying to avoid a large list of modifiers that don't produce a very noticeable effect. Some of the modifiers sometimes can produce just the same image every time, in which case it won't be generally useful for different prompts.

Please confirm that each modifier being added (a) creates the desired effect, (b) doesn't produce the same or almost-same image regardless of the prompt.

Thanks! I appreciate the contribution! :)

MrManny:
I didn't want to open a PR for just this; but I noticed "Bob Ross" isn't in the list of modifiers 😅
Is it possible to throw him in there in this PR?

thomasbaijot:
Oh, I didn't do that 😵‍💫
Now I know how to contribute on GitHub, I will take my time to check every
modifier and clean and organize the list.
Is it possible to create sections in sections ?
I.e :
- Artists
     - Painters
     - Illustrators
     - Sculptors
     - Companies
- Moods
     - Positive moods
     - Negative moods
...

PR #198, opened 2022-09-15
Update style of buttons.

mrbusysky:
there were looking a little old. So i made them match everything else.


cmdr2:
Thanks! I've added these changes separately, because this PR is against an outdated version of the code (too many conflicts).

PR #233, opened 2022-09-23
Adjust card label size for very small cards

Hakorr:
| Before | After |
|--------|-------|
|    image    |    image   |


cmdr2:
Done, thanks!

PR #234, opened 2022-09-23
How to fix: 'chcp' is not recognized

JeLuF:
Got this error message twice in the #support channel on discord tonight, might help to have it in the troubleshooting guide. Should I add pictures?


cmdr2:
Thanks @JeLuF ! I've actually gone ahead and made a change to the installer to always include `C:\Windows\System32` in the `PATH`. Commit: https://github.com/cmdr2/stable-diffusion-ui/commit/858a1c7ae00e5e16c9d62a7e7fa876f70d2a79cd

So I'll close this for now, because the installer will prevent this situation from happening in the first place.

Incase we continue seeing these errors, I'll re-open and merge your PR. Thanks!

PR #240, opened 2022-09-24
Fix for file name lengh limit on Windows

HPPinata:
This is more reliable and allows for better organisation than just moving the folder to C:. Its also benificial in unrelated scenarios (and honestly should be the default in windows).

After making this change I was able to install both versions i tested (v2.05 and v2.16) in "C:\Users\<USER>\Downloads" which did not work before


samuelbarrionuevo (approved):

PaarthShah (approved):

HPPinata:
This is apparently not getting merged and while trying to resolve the conflict due to the Readme update from .tar.xz -> .zip in the GitHub webui I accidentally closed it.

PR #242, opened 2022-09-25
Image item refactor

Hakorr:
This PR fixes and refactors code related to image items.

### Issues which were fixed

1. The download button goes to the wrong place on small screens.
2. Download filename contains the first image's seed if multiple images are made.

### DIFF
--- a/ui/media/main.css
+++ b/ui/media/main.css
@@ -69,38 +69,28 @@ label {
     font-size: 8pt;
 }
 .imgSeedLabel {
-    position: absolute;
-    transform: translateX(-100%);
-    margin-top: 5pt;
-    margin-left: -5pt;
-    font-size: 10pt;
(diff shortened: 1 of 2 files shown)


cmdr2:
Does this work with Live Preview? I believe the `makeImageContainers` function is used by the streaming updates as well: https://github.com/Hakorr/stable-diffusion-ui/blob/main/ui/media/main.js#L384 (which is now a dangling function). Thanks

Hakorr:
> Does this work with Live Preview? I believe the `makeImageContainers` function is used by the streaming image updates as well: https://github.com/Hakorr/stable-diffusion-ui/blob/main/ui/media/main.js#L384 (which is now a dangling function).

Ouch, totally missed live previews. Have to check that out.

PR #284, opened 2022-10-05
Restrict UI to localhost

Teemu:
It might not be wise to have the Stable Diffusion UI running on public Internet considering it has the ability to create directories & save random photos.

This restricts the access to localhost only.

### DIFF
--- a/scripts/on_sd_start.bat
+++ b/scripts/on_sd_start.bat
@@ -312,6 +312,6 @@ call WHERE uvicorn > .tmp
 
 @call python --version
 
-@uvicorn server:app --app-dir "%SD_UI_PATH%" --port 9000 --host 0.0.0.0
+@uvicorn server:app --app-dir "%SD_UI_PATH%" --port 9000 --host 127.0.0.1
 
 @pause
--- a/scripts/on_sd_start.sh
+++ b/scripts/on_sd_start.sh
@@ -304,6 +304,6 @@ cd stable-diffusion
 
 python --version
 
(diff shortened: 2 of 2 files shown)


xanderfrangos:
I often have Stable Diffusion UI running on my desktop while I access it from another device in my home. I've also, at times, temporarily opened up the port so friends and co-workers could play around with some prompts. If access is restricted to localhost by default (for valid security reasons), maybe there should be an option to open it up again without editing the script?

Teemu:
@xanderfrangos That's certainly a valid use case and I agree it could be an optional setting somewhere. Alternatively, I would be happy if the option to save images locally would be disabled if the request didn't come from localhost.

cmdr2:
Hi, I'm working on having this come from a config.json file, in an upcoming update. Maybe the default there could be localhost instead of 0.0.0.0. It will break the experience for many users (who'll have to change their settings after the update), but I understand the security concern, so I'll probably make the default localhost in the upcoming update.

I'll keep this PR open until the new version (with a configurable host and port) comes out. Thanks

Seedmanc:
Will there be a way to keep the online functionality? I'm currently hosting it for friends, it would be sad to lose that.

patriceac:
How about having proper authentication?

MateKristof:
Maybe a simple password?

cmdr2:
We have a way to configure this in the UI now (in the Settings tab). Keeping this PR open until we change the default to localhost

PR #295, opened 2022-10-07
React base display reorg

caranicas:


ui/sd_internal/runtime.py
```diff
@@ -218,7 +218,9 @@ def do_mk_img(req: Request):
     res.request = req
     res.images = []
 
-    temp_images.clear()
+    # keep the cached images in memory
```
caranicas, line 221: @cmdr2  this change and adjustments to the server where things I needed to do to get the progress images to persist on the front end. without this, each new image was trashing the previous progress images, which resulted in some funky display issues with the browser cache.

PR #301, opened 2022-10-08
Resize canvas following selected aspect ratio

madrang:
first implementation of resizeInpaintingEditor function.

### DIFF
--- a/ui/media/main.js
+++ b/ui/media/main.js
@@ -12,6 +12,7 @@ const SHOW_ONLY_FILTERED_IMAGE_KEY = "showOnlyFilteredImage"
 const STREAM_IMAGE_PROGRESS_KEY = "streamImageProgress"
 const HEALTH_PING_INTERVAL = 5 // seconds
 const MAX_INIT_IMAGE_DIMENSION = 768
+const INPAINTING_EDITOR_SIZE = 400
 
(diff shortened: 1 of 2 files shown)


ui/media/main.js
```diff
@@ -262,6 +263,31 @@ async function healthCheck() {
...
+        widthValue = INPAINTING_EDITOR_SIZE
+        heightValue = INPAINTING_EDITOR_SIZE
+    } else if (widthValue > heightValue) {
+        heightValue = (heightValue / widthValue) * INPAINTING_EDITOR_SIZE
+        widthValue = INPAINTING_EDITOR_SIZE
+    } else {
+        widthValue = (widthValue / heightValue) * INPAINTING_EDITOR_SIZE
+        heightValue = INPAINTING_EDITOR_SIZE
+    }
+    inpaintingEditorContainer.setAttribute("style", `width:${widthValue}px;height:${heightValue}px`)
```
cmdr2, line 279: Why is this line set, if we're changing the style.width and height in the next two lines?

Plus, this line will overwrite any custom styling (which we shouldn't do, and don't do). But is this line really necessary?

madrang: This is to resize the div `inpaintingEditorContainer`
That's that standard way to resize a div if you need to support both Chrome and Firefox.
https://stackoverflow.com/a/10118190

cmdr2:
Merged, thanks! :)

PR #303, opened 2022-10-08
Resize in-painting disable

madrang:
Disable 'resizeInpaintingEditor' when 'maskSetting' is unchecked and run 'resizeInpaintingEditor' once when 'maskSetting' is changed in case the aspect ratio was changed while 'resizeInpaintingEditor' was hidden.


cmdr2:
Merged, thanks!

PR #310, opened 2022-10-08
async ping locks on load_model_ckpt, avoid lock at start by preloading model before browser open

madrang:
`async def ping():` will lock on `load_model_ckpt` as the method loading the model is not async.
To avoid the lock at the start we could preload the model before calling browser open.

[Src](https://dev.to/ruarfff/understanding-python-async-with-fastapi-42hn): `If you don't specify your endpoint as being async i.e. as [a coroutine](https://docs.python.org/3/library/asyncio-task.html), FastAPI will look after it for you instead as it will assume this route contains blocking calls whereas if you specify async, it's up to you to make sure it really is asynchronous.
`

### DIFF
--- a/ui/server.py
+++ b/ui/server.py
@@ -333,6 +333,8 @@ async def start_browser():
     # ping once to load the model before opening the browser
     # avoid opening the browser when the FastApp server is locked loading a new model
     await ping()
+    # added sleep to avoid intermitent timeout error.
+    await asyncio.sleep(1)
     # start the browser ui
     import webbrowser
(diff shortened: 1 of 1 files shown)


madrang:
When testing on my PC, when `load_model_ckpt` is being called, the web server will not respond.
This tries to avoid this a little by waiting until the first model is loaded before opening the web ui.

madrang:
Will reimplement ping to not lock server at all. Closed as there is a better way to fix this.

Issue #306, opened 2022-10-08
Save to PNG  by default, instead JPEG


maikelsz:
PNG is higher quality. PNG is **lossless** even when uses some compression (the compression parameter goes to 9, but compressing a PNG is slower on complex images, cartoons/anime compress very well).

In some JPEG implementations/libraries you can disable down-sampling of the chroma channels, to avoid losing vibrancy and color detail. In addition, set the default quality to at least level 89. higher values are not very noticeable, (on flatter images you con notice the artifacts on the edges and hard gradients) but the file size increases almost quadratic or exponentially. I personally use 100% for some jobs, and so it is still smaller than a PNG (depend of the image content), although always with loss. but for files I keep as a backup that will no longer be modified, I still set 89)

towcar:
Should we not be switching to a more modern file format like webp?

maikelsz:
> Should we not be switching to a more modern file format like webp?

I've seen many editors and image viewers strugle with Wep files (which is why I personally avoid it as much as I can, and if I have Wep files I convert them to JPEG with as little difference/loss as possible, and then restore any artifacts that may Appear)

the use of PNG allows to have a lossless format (even its compression is lossless) with universal support.

JPEG also has a lossless variant, but it would miss the point of using jpeg

Wep also has the ability to save compressed without loss, but then the same thing happens to it as with jpeg

JeLuF:
In the system settings, enable "Automatically save settings", then click on configure and check whether the Output Format checkbox is checked. If you now switch to PNG, all future images will be saved as PNG.

maikelsz:
yeah, I forgot I saw that latter. something is something

PR #314, opened 2022-10-10
change default jpeg -> png

lidarbtc:
#306 
png default, instead jpeg.


cmdr2:
Hi, What's the reason to switch the default to PNG? The UI remembers your preference, so if you want to keep using PNG, just set that once. Subsequent requests will remember to use PNG.

A change in beta (will soon release to main) will preserve the output format preference across restarts.

Given that upcoming change, I'd prefer to keep the default as JPEG. If someone really wants the lossless nature PNG, they can set that preference once in their UI, and it'll remember across restarts.

Otherwise, the visual difference isn't enough to justify the large file sizes of PNG, especially for users who're making large batches of images.

PR #329, opened 2022-10-14
Task queue rendering

madrang:


ui/server.py
```diff
@@ -307,14 +486,21 @@ def getModels():
...
 
-@app.get('/modifiers.json')
-def read_modifiers():
-    headers = {"Cache-Control": "no-cache, no-store, must-revalidate", "Pragma": "no-cache", "Expires": "0"}
-    return FileResponse(os.path.join(SD_UI_DIR, 'modifiers.json'), headers=headers)
-
-@app.get('/output_dir')
-def read_home_dir():
-    return {outpath}
+@app.get('/get')
```
cmdr2, line 489: How about `/get/{key}`, so that it follows a RESTful pattern?

ui/server.py
```diff
@@ -65,38 +92,152 @@ class ImageRequest(BaseModel):
...
+        current_state_error = None
+        current_state = ServerStates.Online
     except Exception as e:
+        current_model_path = None
+        current_state_error = e
+        current_state = ServerStates.Unavailable
         print(traceback.format_exc())
-        return HTTPException(status_code=500, detail=str(e))
+
+def thread_render():
```
cmdr2, line 164: Can we move all the task-related classes and functions to another file? Maybe something like `task_manager.py`? So we'd have 3 files: `server.py` (handles the server interface), `task_manager.py` (manages the tasks and queue), `runtime.py` (interacts with the underlying SD engine. Maybe we should call it `renderer.py`?)

ui/server.py
```diff
@@ -242,42 +436,27 @@ async def setAppConfig(req : SetAppConfigRequest):
         print(traceback.format_exc())
         return HTTPException(status_code=500, detail=str(e))
 
-@app.get('/app_config')
-def getAppConfig():
+def getConfig(default_val={}):
```
cmdr2, line 439: Nice, thanks for the cleanup here :)

ui/server.py
```diff
@@ -307,14 +486,21 @@ def getModels():
...
-@app.get('/modifiers.json')
-def read_modifiers():
-    headers = {"Cache-Control": "no-cache, no-store, must-revalidate", "Pragma": "no-cache", "Expires": "0"}
-    return FileResponse(os.path.join(SD_UI_DIR, 'modifiers.json'), headers=headers)
-
-@app.get('/output_dir')
-def read_home_dir():
-    return {outpath}
+@app.get('/get')
+def read_web_data(key:str=None):
```
cmdr2, line 490: Nice idea!

cmdr2 (commented):
Read through quickly, looks good at a broad level. I haven't gone into the details of how the task management code works in server.py, or the details of the logic flow in main.js.

But it looks good broadly. And like you already know, we'll definitely need to test this a lot :)

Nice work!

ui/server.py
madrang: I used query to make sure that for replies where we don't specify a header, that the result won't be cached.
Mosts browsers will explicitly disable cache for url with a query.
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/QueryStringParameters.html

ui/server.py
madrang: We could for sure, but I didn't want to add one more file to distribute.
At this point it's just cleaning up a little and should be easy now that it's working.

ui/server.py
cmdr2: We call send a cache header fairly easily (see the cache header sent in another app path). For now, sending a "never cache" header is correct, since we're sending live config data like models and app config, which should always be fresh

ui/server.py
cmdr2: Adding another file is not a problem. The auto-updater copies the entire UI folder over, so we don't need to do anything extra wrt that.

So this would mean:
1. add `sd_internal/task_manager.py`
2. in `server.py`, use `from sd_internal import task_manager`, and update the references in the code to `task_manager.foo_function()` instead of `foo_function()`
3. in `task_manager.py` use `import .runtime` incase you need to interact with the runtime.

I'd definitely definitely prefer to keep these separate, and move the task-related code into `task_manager.py`. I'm trying to avoid adding additional bloat to already-bloated files, and am working on gradually refactoring the code (JS and Python). I gave a similar comment to the JS-related PR submitted earlier today by another developer.

Thanks!

ui/media/main.js
```diff
@@ -277,16 +299,40 @@ function playSound() {
 
 async function healthCheck() {
     try {
-        let res = await fetch('/ping')
-        res = await res.json()
-
-        if (res[0] == 'OK') {
-            setStatus('server', 'online', 'success')
+        let res = undefined
+        if (sessionId) {
```
cmdr2, line 303: Will this ever be false? Do we need this `if` since we generate the `sessionId` at the top?

PR #334, opened 2022-10-14
CSS styling update and auto-saving settings

mdiller:
I was intending to just clean up some css styling stuff but then had some other ideas and got a bit carried away. Feel free to let me know if any of the things listed below should be changed/removed.

image

# Changes
- Updated the "Image Settings" input form section to make it more readable
- Updated the style of all of the input controls to match
- Darkened the style of the input controls
  - Note that i left checkboxes, silders, and buttons alone for the most part. I could update the styling for these too, but it'd be more involved and im not sure what people would want.
  - Also updated the colors for checkboxes and sliders to use the purple we're using for the "Make Image" button
- Added/used the "Work Sans" font
- Added some additional styling to the image uploaded img2img
  - Added rounded corners 
  - It will now will show the aspect ratio correctly
  - Added a little note to the bottom left of the image that displays the "width x height"
  - If you hover on the image for more than 0.5 seconds, it will grow larger (for easier comparison to the generated results)
- Added support for auto-saving settings. (prompt + things in the "Image Settings" section)
  - Turning on/off this feature is doable from the system-settings section
    - A checkbox is there to disable/enable the feature
    - A button is there to reset settings to default
    - Another button is there to bring up the configuration
  - You can bring up a window to configure which settings are persisted
  - Currently the behavior is really that it always saves and just doesnt *load* if the user hasn't enabled the feature. I figured that'd result in less confusing behavior. Let me know if you want me to change that to not save anything unless the user checks the box.
  - The way its implemented, it should be really easy to add new settings to this if new ones get added in the future, just add it to the `SETTINGS_TO_SAVE` list and it should get added smoothly
  - It currently doesn't support auto-saving the image loaded for img2img or any of the "Image Modifiers", but it would be possible to add support for that if its wanted
  - I could also at some point add support for saving different presets for settings if that is wanted

I'm happy to revert/remove any of the changes in the above list, I figured some of them like the dark controls or the font might not be desired. Just let me know if we want to keep some of them but not all of them and I can update this PR. If we want dark/light mode but want the option to swap between them let me know and i can implement that too. I can also split this up into 2 PRs (one for the visual updates and one for the auto-saving settings) if you want that instead.

Let me know if you want anything changed!

## Images
Heres an image that shows the majority of the changes with a comparison to the old version

image

Click on any of the below things to un-hide the images
<details>
  <summary>Without the dark coloring</summary>

image

</details>
<details>
  <summary>After hovering over the img2img image for 0.5s</summary>

image

</details>
<details>
  <summary>The new entry in the system settings</summary>
Unchecked:

image

Checked:

image

</details>
<details>
  <summary>The configuration screen that comes up when you hit the "Configure" button</summary>

image

</details>

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -8,6 +8,7 @@
     <link rel="stylesheet" href="/media/modifier-thumbnails.css?v=1">
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.2.0/css/all.min.css">
     <link rel="stylesheet" href="/media/drawingboard.min.css">
+    <link href="https://fonts.googleapis.com/css2?family=Work+Sans:wght@400;600;700;800&display=swap" rel="stylesheet"> 
     <script src="/media/jquery-3.6.1.min.js"></script>
     <script src="/media/drawingboard.min.js"></script>
 </head>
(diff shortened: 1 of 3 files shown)


mdiller:
Closing. Pull request recreated as #335 at the suggestion of **Madrang** in discord

PR #336, opened 2022-10-15
Downloaded images named with their correct prompt and settings

ZacTheHac:
I was originally writing a bug report, but felt this was more helpful.

Downloading an image sets the filename to [prompt]_Seed-[seed]_Steps-[steps]....etc
The [prompt] section is always whatever the last prompt was instead of the prompt associated with that image generation.
My changes embed the prompt, inference steps, and guidance scale into the imageData object, to be used when generating the file name.

To reproduce the issue this fixes:
Start generating one image, say "Male portrait, Figurative"
Enqueue a second, say "Male portrait, Constructivist"
Click "Download" on the first image
the file name will be  "male_portrait__constructivist", the correct seed, but the steps/guidance will be whatever was last used, not what was used for that image

### DIFF
--- a/ui/media/main.js
+++ b/ui/media/main.js
@@ -273,6 +273,9 @@ function showImages(req, res, outputContainer, livePreview) {
 
         const imageData = result?.data || result?.path + '?t=' + new Date().getTime(),
             imageSeed = result?.seed,
+            imagePrompt = req.prompt,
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks! I really appreciate you fixing this :)

PR #350, opened 2022-10-18
Fixed file path bugs introduced by mistake and made img_id sequential.

madrang:
Fixed file path bugs introduced by mistake and made img_id sequential based on time for better sorting of renders.


madrang:
Results of the change
image

madrang:
The img_id used are now a predictable pattern that then sorted alphabetically will also sort by prompt/time.

PR #351, opened 2022-10-18
Force encoding to utf-8 on text file operations Fixes #332

### LINKED ISSUE #332: Txt files fail to write if there's unicode in the prompt
When there are unicode character like 😄 or ō in a prompt, and you are using the "automatically save to"  option, the image generates properly but the associated text file is created but is empty. Running a headless setup so don't have the exact error message but it should be trivial to replicate.

Err message starts with
Unicode encode error

Steps to replicate 
1. Setup the automatically save to
2. Enter a prompt with a unicode character like ō or 😃 in it
3. inspect file result

OS:  Win 11
Browser: doubt it's a browser issue, but Firefox
Version: v2.22 (beta)

madrang:
Fix for https://github.com/cmdr2/stable-diffusion-ui/issues/332
Force open to use encoding utf-8 for text file operations. Fixes systems where the default is another encoding.

### DIFF
--- a/ui/sd_internal/runtime.py
+++ b/ui/sd_internal/runtime.py
@@ -541,7 +541,7 @@ Negative Prompt: {req.negative_prompt}
 Stable Diffusion Model: {req.use_stable_diffusion_model + '.ckpt'}
 """
     try:
-        with open(meta_out_path, 'w') as f:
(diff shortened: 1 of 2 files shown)


cmdr2:
Thanks!

PR #353, opened 2022-10-18
Theme Selection

mdiller:
Added a way to select themes from the ui. The theme gets saved with the rest of the auto-settings. I added some basic themes, and tried to get some decent examples of how customizable it is. I'm no UX designer but I think some of these look decent. Would definetly be happy to see some additions to the themes by someone who is good at colors.

Also at some point a way to do a "Custom" theme could be cool, where the user can select the colors and watch it change. I gotta go to sleep now though so this is it for me today.

It should be super easy to add new themes, just add them in any included css file as a .theme-example-name and set variables like the ones at the top of main.css. The js code will look for those and auto-populate the dropdown box with them. 

You can select a theme here:
image

Heres the current available themes:
Default
image

Light
image

Discord
image

Cool Blue
image

Blurple
image

Super Dark
image

Wild
image

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -43,6 +43,7 @@
                         <li>
                             <input id="auto_save_settings" name="auto_save_settings" type="checkbox">
                             <label for="auto_save_settings">Automatically save settings <small>(settings restored on browser load)</small></label>
+                            <br/>
                             <button id="configureSettingsSaveBtn">Configure</button>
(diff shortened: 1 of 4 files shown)


cmdr2:
Thanks, merged it in. I'll move the themes code into a separate JS file (maybe `themes.js`) though. It'll be good to avoid bloating `main.js` any more than it is, I'm actively trying to split `main.js` into separate files actually.

Love this change, thanks a bunch!

PR #361, opened 2022-10-19
fix seed for parallel renders with filters applied

rbertus2000:
This fixes the seed output for parallel renders with applied filters


cmdr2:
Thanks!

PR #363, opened 2022-10-20
Settings Refactor

mdiller:
Changes:
- The issue with the guidance scale is fixed
- All other settings have been migrated over to auto-save.js
- Refactored the way the settings work to be a bit cleaner, now just has one variable (`SETTINGS`) to keep track of stuff.
- Settings now load before all the other stuff loads, makes it look much smoother than before
- Settings sorted by section in the "configure" panel
- Added a button for restoring defaults of just the "Image Settings" pannel (figured this is where people would want to be able to reset)
- Refactored the collapsibles stuff so its all in utils.js now, and it auto-finds the .collapsible marked stuff to save on startup (could add more collapsible panels in the future without js changes)
- I noticed you'd removed all the end of line semicolons so i did my best to keep the code that way


mdiller:
oops this is into the wrong branch. closing this.

PR #364, opened 2022-10-20
Settings Refactor

mdiller:
Changes:
- The issue with the guidance scale is fixed
- All other settings have been migrated over to auto-save.js
- Refactored the way the settings work to be a bit cleaner, now just has one variable (`SETTINGS`) to keep track of stuff.
- Settings now load before all the other stuff loads, makes it look much smoother than before
- Settings sorted by section in the "configure" panel
- Added a button for restoring defaults of just the "Image Settings" pannel (figured this is where people would want to be able to reset)
- Refactored the collapsibles stuff so its all in utils.js now, and it auto-finds the .collapsible marked stuff to save on startup (could add more collapsible panels in the future without js changes)
- I noticed you'd removed all the end of line semicolons so i did my best to keep the code that way


ui/index.html
```diff
@@ -8,6 +8,7 @@
     <link rel="stylesheet" href="/media/css/themes.css?v=1">
     <link rel="stylesheet" href="/media/css/auto-save.css?v=1">
     <link rel="stylesheet" href="/media/css/main.css?v=1">
+    <link rel="stylesheet" href="/media/css/auto-save.css?v=1">
```
cmdr2, line 11: Thanks, but this was already added in a bug fix (two lines above this one) :)

ui/index.html
```diff
@@ -104,7 +104,7 @@ <h1>Stable Diffusion UI <small>v2.3.2 <span id="updateBranchLabel"></span></smal
             <div class="line-separator">&nbsp;</div>
 
             <div id="editor-settings" class="panel-box settings-box">
-                <h4 class="collapsible">Image Settings</h4>
+                <h4 class="collapsible">Image Settings<i id="reset-image-settings" class="fa-solid fa-arrow-rotate-left"></i></h4>
```
cmdr2, line 107: Can we have a tooltip displayed over this? Or change it to a text button saying "Restore Defaults"?

cmdr2:
The PR looks pretty good, thanks!

Pasting our discussion from Discord regarding preserving the older settings:

Both kinds of settings:
- the old-style "save to disk", "use CPU", "full precision" etc - these should not be lost
- the "guidance scale", "negative prompts" etc - these should also not be lost after this refactor is merged. People have been using this feature for a few days (a week for beta users), so it'll be good not to lose this.

I realize the new system is cleaner, so maybe a simple script that "migrates" (i.e. copies) the old settings from the previous localStorage key to the new one? We can keep this script around for maybe a month or two, and then delete it.

Sorry about the extra work, but would be good to avoid a negative surprise for users :)

ui/media/js/main.js
```diff
@@ -1146,37 +1031,18 @@ function onDimensionChange() {
...
-if (isNegativePromptPanelOpenEnabled()) {
-    setPanelOpen(negativePromptPanelHandle)
-}
-
 makeImageBtn.addEventListener('click', makeImage)
 
 
 function updateGuidanceScale() {
     guidanceScaleField.value = guidanceScaleSlider.value / 10
+    guidanceScaleField.dispatchEvent(new Event("change"))
```
cmdr2, line 1045: I've fixed this temporarily in beta (didn't want to wait for this full PR to fix the bug). But I like your solution better, so when you're resubmitting this PR, please delete my fix.

I put `guidanceScaleSlider.dispatchEvent(new Event("change"))` in the next function `updateGuidanceScaleSlider()`, since the current auto-save code still listens to guidanceScaleSlider. And similarly for the prompt strength bit as well.

Please remove my fix before submitting yours, since yours is better.

mdiller: I couldn't find your change for this? Not sure if it got removed already or if I missed it.

ui/index.html
mdiller: Gotcha, I've removed my line!

ui/index.html
mdiller: Sure! I've added a tooltip. Not sure how we want to do tooltips so i copied some tooltip css from one of my other projects, and we can change it as we like in the future. If you are looking for some more specific styling for tooltips let me know and i can update it.

mdiller:
No worries at all, having support for supporting the old version of settings makes sense. I've added support for it now. It should work with all previous setting types, and deletes the old settings from `localStorage` after it consumes them. Let me know if theres anything else!

Issue #392, opened 2022-10-26
Resume Downloads


George-D-S:
**Is your feature request related to a problem? Please describe.**
I'm having trouble getting it to install because the larger downloads are failing. My internet connection can be flakey at times.

The first issue was with sd-v1-4.ckpt but I could work round this by getting the URL from the script and do the download manually.

I'm now stuck with it having opened the browser window and downloading something it's putting under

.cache\huggingface\transformers

but I'm unclear of what this is or where I'd get it from.

**Describe the solution you'd like**

I'd like any downloads to have an auto resume-retry option, or for instructions on how to work round it by downloading the files directly.

JeLuF:
I think this would be useful:
`curl -C - --retry 20 --retry-all-errors --retry-delay 5 -L -k https://me.cmdr2.org/stable-diffusion-ui/sd-v1-4.ckpt > sd-v1-4.ckpt`

Risk: "-C -" will use any existing file and append to it - whether it's from the same source or something completely different.
All current calls to curl have an `@if not exist` statement that ensures that this problem will not occur.

JeLuF:
`curl --retry 5` does not retry the download if the connection gets lost. For this, `--retry-all-errors` is needed, wich was added to curl two years ago. According to yesterday's experience, there are a lot of older versions of curl still in use, so we can't rely on `--retry-all-errors` to be available.

1. Download a new version of curl early during the installation and use it for all following downloads.
2. Write a retry loop. Simple in `bash`, ugly in `cmd`. Ideally, using `curl -o`, so that `-C` can be used.

JeLuF:
Draft versions
### CMD
```
@CALL :trycurl 20 "https://me.cmdr2.org/stable-diffusion-ui/sd-v1-4.ckpt" "xxx.data"
@echo Download completed.


@GOTO :EOF
REM ####################################################################################
REM ### Helper subroutines
REM ####################################################################################

REM ------ trycurl -------
REM Try %1 times to download %2 to the file %3
REM The file gets deleted before trying the download
REM The download will resume where it failed before
REM ----------------------
:trycurl
@set count=%1
@set url=%2
@set dest=%3
@del %dest%
:DoWhile
    @echo trying to download (%count% attempts left)
    @if %count%==0 goto EndDoWhile
    @set /a count = %count% -1
    @call curl -C - -L %url% -o %dest%
    @if %errorlevel%==0 exit /b
    @timeout /t 5 /nobreak
[...]

PR #394, opened 2022-10-26
Retry downloads on flaky internet connections

JeLuF:
Issue #392 
Change curl settings to retry up to 20 times to continue a download, resuming where the previous attempt failed, sleep for 5s between retries (so that the 20 retries don't get consumed too quickly).


cmdr2:
Nice, thanks for this! :)

PR #404, opened 2022-10-27
Support for multiple GPUs, and improvements to RAM and VRAM usage

cmdr2:
Credit: @madrang


ui/server.py
```diff
@@ -292,8 +358,56 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    if isinstance(config['render_devices'], str):
+        config['render_devices'] = config['render_devices'].split(',')
+    if not isinstance(config['render_devices'], list):
+        raise Exception('Invalid render_devices value in config.')
+    for device in config['render_devices']:
+        task_manager.start_render_thread(device)
+else:
+    # Select best GPU device using free memory, if more than one device.
+    task_manager.start_render_thread('auto') # Detect best device for renders
+    task_manager.start_render_thread('cpu') # Allow CPU to be used for renders
```
rbertus2000, line 407: Is there any reason, why you start loading the model to gpu and cpu for "new" users without render_devices config?

madrang: It's allowing them to be used, but preload checks and will only work on gpu. Also tonight a change is coming to allow preload on cpu when no gpu are present. But at the moment preload is not applied on cpu.

cmdr2: I've pinged @madrang regarding this question

madrang: But Cmdr2 did change it ti always preload...
https://github.com/cmdr2/stable-diffusion-ui/pull/404/commits/389e3397ecc8b2e47217d0428a8291d0c5253e9b

cmdr2: Thanks @rbertus2000 for pointing this out! This bit will undergo some changes soon

PR #417, opened 2022-10-29
Moved System Settings & Reworked into "Parameters"

mdiller:
Changes
- updated the "popup" system to be more generic and improved its behavior a bit
- moved the system settings to a dialog box
- implemented the system settings as "parameters"

About the new "Parameters"
- I didn't call them settings because we already have "settings" meaning something else (auto-save stuff that includes "Image Settings")
- parameters is maybe a weird name and I'm definitely open to change it or however you want
- I created a new file for them called "parameters.js"
- we could do a lot more with the system like support auto-hide/disable of other parameters etc, but I kept it pretty simple for now
- I defined the system settings parameters in JS but we should be able to define other parameters as json for plugins etc. One advantage of doing it in JS is we can get some intellisense with the JSDoc stuff.
- this is a bit proof-of-concept-y because I'm not sure where we want to take the system and how advanced/complicated it should be
- one example of a way we could make this better/more advanced is have a universal way to sign up to change events on a parameter/setting

Oh also I left the "Help & Community" stuff how it is for the moment, I'm not sure what you wanna do with it.

**Screenshots:**
Before:
image

After:
image

**Other:**
<details>
<summary>Before updating the labels/text</summary>

image

</details>

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -240,9 +240,9 @@
         </div>
     </div>
 
-    <div id="save-settings-config" style="display:none">
+    <div id="save-settings-config" class="popup">
         <div>
-            <span id="save-settings-config-close-btn">X</span>
+            <i class="close-button fa-solid fa-xmark"></i>
             <h1>Save Settings Configuration</h1>
             <p>Select which settings should be remembered when restarting the browser</p>
             <table id="save-settings-config-table">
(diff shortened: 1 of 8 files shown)


cmdr2:
Really nice work! Thanks :)

PR #421, opened 2022-10-29
Fix plugins needing to specify many params or they would be missing in the render request.

madrang:
Don't reset reqBody, only replace using req as we use a new task object created from UI inputs.
Fix plugins needing to specify many params or they would be missing in the render request.


cmdr2:
Thanks!

PR #422, opened 2022-10-29
Implement complete device selection in the backend.

madrang:
Also fix having to specify cpu when it's the only device present.
use_cpu still works but maps to render_device: 'cpu'

### DIFF
--- a/ui/sd_internal/task_manager.py
+++ b/ui/sd_internal/task_manager.py
@@ -38,6 +38,7 @@ class RenderTask(): # Task with output queue and completion lock.
     def __init__(self, req: Request):
         self.request: Request = req # Initial Request
         self.response: Any = None # Copy of the last reponse
+        self.render_device = None
         self.temp_images:list = [None] * req.num_outputs * (1 if req.show_only_filtered_image else 2)
         self.error: Exception = None
         self.lock: threading.Lock = threading.Lock() # Locks at task start and unlocks when task is completed
@@ -68,7 +69,8 @@ class ImageRequest(BaseModel):
     # allow_nsfw: bool = False
     save_to_disk_path: str = None
     turbo: bool = True
-    use_cpu: bool = False
+    use_cpu: bool = False ##TODO Remove after UI and plugins transition.
+    render_device: str = None
     use_full_precision: bool = False
     use_face_correction: str = None # or "GFPGANv1.3"
     use_upscale: str = None # or "RealESRGAN_x4plus" or "RealESRGAN_x4plus_anime_6B"
@@ -89,7 +91,7 @@ class FilterRequest(BaseModel):
     height: int = 512
     save_to_disk_path: str = None
     turbo: bool = True
-    use_cpu: bool = False
+    render_device: str = None
     use_full_precision: bool = False
     output_format: str = "jpeg" # or "png"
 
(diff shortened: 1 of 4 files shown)


ui/sd_internal/task_manager.py
```diff
@@ -268,6 +273,11 @@ def thread_render(device):
             return
         task = thread_get_next_task()
         if task is None:
+            if runtime.thread_data.device == 'cpu' and is_alive() > 1 and hasattr(runtime.thread_data, 'lastActive') and time.time() - runtime.thread_data.lastActive > CPU_UNLOAD_TIMEOUT:
+                # GPUs present and CPU is idle. Unload resources.
+                runtime.unload_models()
```
cmdr2, line 278: What happens if a user requests 'CPU' again in a subsequent request?

ui/sd_internal/task_manager.py
```diff
@@ -219,26 +222,24 @@ def thread_get_next_task():
...
+                if queued_task.render_device == 'cpu':
                     queued_task.error = Exception('Cpu cannot be used to run this task. Remove GFPGANer filter to run task.')
                     task = queued_task
                     break
                 if not runtime.is_first_cuda_device(runtime.thread_data.device):
                     continue  # Wait for cuda:0
-            if queued_task.request.use_cpu and runtime.thread_data.device != 'cpu':
-                if is_alive('cpu') > 0:
-                    continue  # CPU Tasks, Skip GPU device
+            if queued_task.render_device and runtime.thread_data.device != queued_task.render_device:
```
cmdr2, line 231: In the new logic, it looks like if the program started with a valid GPU (i.e. no thread started for CPU), the user will never be able to switch to CPU in a subsequent request.

It wouldn't throw an error previously, but now it will because `is_alive('cpu')` will be 0 (no thread was started for at the start).

Apologies if I've misunderstood how this works.

ui/sd_internal/task_manager.py
```diff
@@ -341,6 +354,21 @@ def get_cached_task(session_id:str, update_ttl:bool=False):
         return None
     return task_cache.tryGet(session_id)
 
+def get_devices():
```
cmdr2, line 357: My intention behind a UI-based device selection was to list *all* available devices (GPUs and CPU), not just the ones we picked up from config.json, or created render threads for.

The intent is to write the chosen devices to config.json. I'd like to avoid the user needing to edit config.json manually for device selection.

This kind of UI-device device selection will improve the adoption of this feature (visible in the UI), and make it simpler for users to use this. It's friendlier to be presented with a list of devices available on the PC, and choose one or multiple, and have that be written to config.json automatically for them.

cmdr2:
Thanks @madrang , it's nice how this PR unifies the device-related code! Nice work :)

I'm merging this PR, but could you please check the edge cases I pointed out (potentially missing edge-cases, but I might be wrong)?

The getDevices() logic is not really what I was talking about. It's okay, but IMO it still falls short and hides this useful feature behind an 'edit-this-JSON-file' barrier. We could argue that it's just a JSON file for people to edit, but that's basically what our UI is all about (vs auto or other UIs) - low-fuss and low-technical setup. Thanks!

ui/sd_internal/task_manager.py
madrang: Noting as long as it's within 4 minutes.
lastActive update each time a new task is sent to it.
So models unloading should not trigger unless CPU is alone and idle for more than CPU_UNLOAD_TIMEOUT

ui/sd_internal/task_manager.py
madrang: It's a start to be able to select one of the present and active device instead of only CPU in the render path.
But it is true that this should be extended to list all device with it's current status (ready/disabled) and allow to start/stop render threads. But this base is needed to implement the rest and should be tested first.

It's just a matter of extending the current method to include more informations and allow more control.
Ill do this over time in another PR.

ui/sd_internal/task_manager.py
cmdr2: What I mean is, if the user makes a request after the model has unloaded (takes a lunch break and comes back)? Won't it just crash?

ui/sd_internal/task_manager.py
cmdr2: Thanks, agreed. I think we're pretty close. Just need the getDevices() to return the CPU plus loop over `torch.cuda.device_count()` to get the GPU names.

That'll unblock us to make the UI change to select different combinations of devices, and have it be saved into config.json.

This will be useful even if it needs a program restart (to implement quickly, without starting/stopping threads automatically).

cmdr2: And apologies for my comment. I'm not expecting or asking you to build the full getDevices() method, and I'm genuinely happy with this start. I just wanted to put my thoughts across on where we need to build this further, not asking you to build it. All of what we do here is totally voluntary.

PR #426, opened 2022-10-30
Improve UI on mobile devices

ayunami2000:
No longer scrolls horizontally on small screens / mobile devices, making it much cleaner to use.

image


cmdr2:
Thanks! :)

PR #427, opened 2022-10-31
Add support for drag&drop for the text files made by the backend

madrang:
Also supports JSON.

### DIFF
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -369,7 +369,7 @@ function checkWriteToClipboardPermission (result) {
         copyIcon.innerHTML = `<span class="simple-tooltip right">Copy Image Settings</span>`
         copyIcon.addEventListener('click', (event) => {
             event.stopPropagation()
-            navigator.clipboard.writeText(JSON.stringify(readUI()))
+            navigator.clipboard.writeText(JSON.stringify(readUI(), undefined, 4))
         })
         resetSettings.parentNode.insertBefore(copyIcon, resetSettings)
     }
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -17,14 +17,22 @@ const TASK_MAPPING = {
     },
     width: { name: 'Width',
         setUI: (width) => {
+            const oldVal = widthField.value
             widthField.value = width
+            if (!widthField.value) {
+                widthField.value = oldVal
+            }
         },
         readUI: () => parseInt(widthField.value),
         parse: (val) => parseInt(val)
     },
     height: { name: 'Height',
         setUI: (height) => {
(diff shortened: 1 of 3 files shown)


ui/media/js/dnd.js
```diff
@@ -0,0 +1,382 @@
...
+    num_inference_steps: { name: 'Steps',
+        setUI: (num_inference_steps) => {
+            numInferenceStepsField.value = num_inference_steps
+        },
+        readUI: () => parseInt(numInferenceStepsField.value),
+        parse: (val) => parseInt(val)
+    },
+    guidance_scale: { name: 'Guidance Scale',
+        setUI: (guidance_scale) => {
+            guidanceScaleField.value = guidance_scale
```
cmdr2, line 61: Will this automatically update the slider as well? Basically, if it automatically triggers the `change` event, then the slider will get updated too. Otherwise we'll have to set it ourselves.

Same for prompt weight.

ui/media/js/dnd.js
```diff
@@ -0,0 +1,382 @@
...
+    use_stable_diffusion_model: { name: 'Stable Diffusion model',
+        setUI: (use_stable_diffusion_model) => {
+            const oldVal = stableDiffusionModelField.value
+
+            const pathIdx = use_stable_diffusion_model.lastIndexOf('/')
+            if (pathIdx >= 0) {
+                use_stable_diffusion_model = use_stable_diffusion_model.slice(pathIdx + 1)
+            }
+            if (use_stable_diffusion_model.endsWith('.ckpt')) {
+                use_stable_diffusion_model = use_stable_diffusion_model.slice(0, use_stable_diffusion_model.length - 5)
```
cmdr2, line 121: `5` => `'.ckpt'.length` please

ui/media/js/dnd.js
madrang: Fixed too by calling updateGuidanceScaleSlider & updatePromptStrengthSlider

ui/media/js/dnd.js
```diff
@@ -0,0 +1,385 @@
...
+
+document.addEventListener("drop", dropHandler)
+document.addEventListener("dragover", dragOverHandler)
+
+// Adds a copy icon if the browser grants permission to write to clipboard.
+function checkWriteToClipboardPermission (result) {
+    if (result.state == "granted" || result.state == "prompt") {
+        const style = document.createElement('style');
+        style.textContent = `
+        #copy-image-settings {
```
cmdr2, line 354: Probably better to keep this in the CSS anyway. If we ever need to style this, it'll be unintuitive for the developer to look in the JS file to find where the 'copy' button is styled.

cmdr2:
Merged, thanks! Nice feature! :)

PR #431, opened 2022-11-01
Copy CUDA_VISIBLE_DEVICES to config.*, it it has been set

JeLuF:
If the CUDA_VISIBLE_DEVICES is already set in config.*, we should copy it over when writing a new config.*, e.g. when switching to beta.


cmdr2:
Thanks! :)

PR #432, opened 2022-11-02
NSIS Installer

JeLuF:
### WORK IN PROGRESS - Probably not yet ready for merging

Scripts to create a Windows installer using Nullsoft's Scriptable Install System.
https://nsis.sourceforge.io/Main_Page
image

### WHY:
- Prevent users from starting Start Stable Diffusion UI.cmd directly from the ZIP file.
- Ensure that the install dir's path isn't too long, doesn't have spaces, and has enough free disk space
- Easy way to have user interaction in the install process. (Alternative: Python-installer with tkinter, but wouldn't solve the above issues)
- Start menu integration 
image

### Future:
- Select model to be installed
image
(currently disabled)
- Bundle curl.exe in a recent version


JeLuF:
Enable Win10 long path name support: https://github.com/conda/constructor/blob/main/constructor/nsis/main.nsi.tmpl#L766

JeLuF:
- Only install on NTFS (FAT installations fail)
- Ask the user to install the Media Feature Pack on Windows 10 N (fixes python cv2 errors)

cmdr2:
Thanks, is this PR okay to merge (you mentioned you wanted to run a few more tests)?

PR #435, opened 2022-11-02
Sanitize session id's before using them as path components

JeLuF:
Session IDs can be generated by the user, so they should be sanitized to ensure that they don't include problematic characters, e.g. \ or /


cmdr2:
Thanks!

PR #438, opened 2022-11-03
Add Use Settings button

patriceac:
Adding Use Settings button (does not restore system settings).

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -589,6 +589,19 @@ input::file-selector-button {
     position: relative;
     transform: translateY(-13%);
 }
+ #copy-image-settings {
+    cursor: pointer;
+    padding: 8px;
+    opacity: 1;
+    transition: opacity 0.5;
+}
+.collapsible:not(.active) #copy-image-settings {
+    display: none;
+}
+#copy-image-settings.hidden {
+    opacity: 0;
+    pointer-events: none;
+}
 
 .simple-tooltip {
 	border-radius: 3px;
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -311,6 +311,8 @@ img {
     float: left;
     display: inline;
     padding-left: 20pt;
+}
+#top-nav-items > li:first-child {
     cursor: default;
 }
 #initial-text {
@@ -322,20 +324,6 @@ img {
     font-weight: bold;
 }
 .pl-5 {
-    padding-left: 5pt;
-}
-#system-settings {
-    width: 360pt;
-    transform: translateX(-100%) translateX(70pt);
-
-    padding-top: 10pt;
-    padding-bottom: 10pt;
-}
-#system-settings ul {
-    margin: 0;
-    padding: 0;
-}
-#system-settings li {
     padding-left: 5pt;
 }
 #community-links {
@@ -677,4 +665,47 @@ input::file-selector-button {
 @keyframes progress-anim { 
     0%   { background-position: -55px 0; }
     100% { background-position: 0 0; }
+}
+
+.popup:not(.active) {
+    visibility: hidden;
+    opacity: 0;
+}
+
+.popup {
+    position: absolute;
+    background: rgba(32, 33, 36, 50%);
+    top: 0px;
+    left: 0px;
+    right: 0px;
+    bottom: 0px;
+    z-index: 1000;
+    opacity: 1;
+    transition: 0s visibility, 0.3s opacity;
+}
+
+.popup > div {
+    position: relative;
+    background: var(--background-color2);
+    border: solid 1px var(--background-color3);
+    max-width: 700px;
+    margin: auto;
+    margin-top: 50px;
+    border-radius: 6px;
+    padding: 30px;
+    text-align: center;
+    box-shadow: 0px 0px 30px black;
+}
+
+.popup .close-button {
+    position: absolute;
+    right: 0px;
+    top: 0px;
+    transform: scale(150%);
+    cursor: pointer;
+    padding: 24px;
+}
+
+#settings-button {
+    cursor: pointer;
 }
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -559,6 +559,26 @@ input::file-selector-button {
         left: 0px;
         right: 0px;
     }
+    .popup > div {
+        padding-left: 5px !important;
+        padding-right: 5px !important;
+    }
+    .popup > div input, .popup > div select {
+        max-width: 40vw;
+    }
+    .popup .close-button {
+        padding: 0px !important;
+        margin: 24px !important;
+    }
+    #reset-image-settings .simple-tooltip.right {
+        right: initial;
+        left: 0px;
+        top: 50%;
+        transform: translate(calc(-100% + 15%), -50%);
+    }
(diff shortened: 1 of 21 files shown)


ui/media/js/image-modifiers.js
```diff
@@ -283,4 +283,13 @@ function loadCustomModifiers() {
     }
 }
 
+// Empty the modifiers when reusing existing settings to avoid duplicate modifiers.
+// Restoring the modifiers will require more in-depth changes. This is a quick workaround for now.
+function clearModifiers() {
+	activeTags.forEach(tag => {
+		tag.originElement.classList.remove(activeCardClass)
+	})
+	activeTags.length = 0
```
cmdr2, line 292: How about `activeTags = []`? `.length` could behave oddly: https://www.tutorialspoint.com/in-javascript-how-to-empty-an-array

ui/media/js/main.js
```diff
@@ -777,7 +785,9 @@ function getCurrentUserRequest() {
             stream_progress_updates: true,
             stream_image_progress: (numOutputsTotal > 50 ? false : streamImageProgressField.checked),
             show_only_filtered_image: showOnlyFilteredImageField.checked,
-            output_format: outputFormatField.value
+            output_format: outputFormatField.value,
+            use_face_correction: useFaceCorrectionField.checked,
```
cmdr2, line 789: Why are these lines being added here, if they're already being added in line 810?

ui/media/js/main.js
```diff
@@ -712,6 +718,8 @@ async function checkTasks() {
 
     task.isProcessing = false
     task['stopTask'].innerHTML = '<i class="fa-solid fa-trash-can"></i> Remove'
+    task['useSettings'].innerHTML = '<i class="fa-solid fa-redo"></i> Use Settings'
```
cmdr2, line 721: If this button's label will always be "Use Settings" (i.e. it never changes), we probably don't need this line right? Because the label is already set in line 853

ui/media/js/main.js
```diff
@@ -839,7 +849,10 @@ function createTask(task) {
     let taskEntry = document.createElement('div')
     taskEntry.className = 'imageTaskContainer'
     taskEntry.innerHTML = ` <div class="taskStatusLabel">Enqueued</div>
-                            <button class="secondaryButton stopTask"><i class="fa-solid fa-trash-can"></i> Remove</button>
+                            <div class="taskAction">
+                                <button class="secondaryButton useSettings" style="display: none"><i class="fa-solid fa-redo"></i>  Use Settings</button>
```
cmdr2, line 853: Do we need the `display: none`? Is there ever a situation where this button should be hidden?

ui/media/js/main.js
```diff
@@ -712,6 +718,8 @@ async function checkTasks() {
 
     task.isProcessing = false
     task['stopTask'].innerHTML = '<i class="fa-solid fa-trash-can"></i> Remove'
+    task['useSettings'].innerHTML = '<i class="fa-solid fa-redo"></i> Use Settings'
+    task['useSettings'].style.display = 'block'
```
cmdr2, line 722: Same question as below - do we really need this line? Is there ever a situation where this button needs to be hidden? If not, we can remove the `display: none` and this `display: block` line, since it'll be visible by default

ui/media/js/main.js
```diff
@@ -876,6 +889,11 @@ function createTask(task) {
         }
     })
 
+    task['useSettings'].addEventListener('click', async function() {
+        TASK_REQ_NO_EXPORT.forEach((key) => delete task.reqBody[key]) // don't restore system settings when restoring tasks
+		restoreTaskToUI(task)
```
cmdr2, line 894: Please use spaces instead of tabs for indentation :)

ui/media/js/main.js
```diff
@@ -683,6 +683,12 @@ async function checkTasks() {
 
     const genSeeds = Boolean(typeof task.reqBody.seed !== 'number' || (task.reqBody.seed === task.seed && task.numOutputsTotal > 1))
     const startSeed = task.reqBody.seed || task.seed
+
+	// update the seed *before* starting the processing, so interrupting the processing retains the latest seed
+    if (randomSeedField.checked) {
```
cmdr2, line 688: Sounds okay. Can you please remove the previous code (now a duplicate) from line 740?

ui/media/css/main.css
```diff
@@ -400,6 +409,17 @@ img {
 .secondaryButton:hover {
     background: rgb(177, 27, 0);
 }
+.useSettings {
+	background: var(--accent-color);
```
cmdr2, line 413: Please use spaces instead of tabs :)

ui/media/js/image-modifiers.js
patriceac: I removed this code and now have proper support for restoring the Modifiers in the left pane.

ui/media/js/main.js
patriceac: Because the existing code at line 810 only works when the boxes are checked, but we also want to actively capture the unchecked state to be able to restore it (otherwise it will just be skipped by TASK_MAPPING). On top of that, I hadn't noticed this code was even there in the first place (I thought these two were just forgotten and added it).

Now that you brought it to my attention, I realized it's actually worse than that- this code is trying to do two things at the same time: capture the state of the checkbox *and* the face correction/upscaling method. I removed the 810 code and refined my addition to better follow the initial implementation's intent while also addressing the box unchecked scenario:
            use_face_correction: useFaceCorrectionField.checked,
            face_correction_method: 'GFPGANv1.3', // hardcoded for now
            use_upscale: useUpscalingField.checked,
            upscale_model: upscaleModelField.value,

Moreover, the init_image and save_to_disk_path calculations just above suffer from the very same issue and currently won't restore the unchecked state because they're just ignoring it.

If we're aligned on the overall direction, I can fix these too while I'm at it.

Thoughts?

ui/media/js/main.js
patriceac: For Use Setting, my initial idea was to only show it once the task is completed. On second thought, I think I overdid it and should just keep it always visible, which also addresses your other two related comments.

Would that be better?

ui/media/js/main.js
patriceac: Yeah, I noticed that and changed the tab behavior in my text editor since then. i cleaned up all my tab tab leftovers and them some.

ui/media/js/main.js
patriceac: Done.

ui/media/css/main.css
patriceac: Yeah, I noticed that and changed the tab behavior in my text editor since then. I cleaned up all my tab tab leftovers and then some, especially in this file. :)

ui/media/js/main.js
patriceac: Made the change. Use Settings is now always visible.

ui/media/js/main.js
patriceac: Made the change. Use Settings is now always visible.

ui/media/js/main.js
patriceac: Made the change. Use Settings is now always visible.

patriceac:
Will redo the whole thing.

PR #453, opened 2022-11-08
Fix modifier and system settings popup position

patriceac:
This fix keeps the settings popup at the same visible position upon scrolling.

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -706,7 +706,7 @@ input::file-selector-button {
 }
 
 .popup {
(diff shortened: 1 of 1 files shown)


cmdr2:
Hmm, what about mobile? Also, I think it needs one more change: `#save-settings-config { position: absolute; }`

Otherwise the `Configure` button inside settings will spill out and not be scrollable. A better design of course would be a max height for the settings box, and use `overflow-y: scroll` for the contents

cmdr2:
Thanks!

PR #455, opened 2022-11-08
Fix: Uncaught TypeError: Cannot set properties of null

JeLuF:
The collapsible render results don't have a handle (+/-) like the image settings, so `handle` can be null.
This has no direct user impact, but causes errors to be logged to the console.


cmdr2:
Thanks!

PR #459, opened 2022-11-09
Fix undefined BIND_IP assigns port instead of ip

rxmzes:
For the Windows on_sd_start batch file: 

The variables added for SD_UI_BIND_IP and PORT were not being assigned properly when undefined, leading to possibly having an invalid port and undefined IP


rxmzes:
Fixed in 65587536ab8771eebba0f7c228bc07615391f425, thanks

PR #461, opened 2022-11-10
Add event listener beforeunload

JeLuF:
When closing the window, a warning is shown if there are any render results.


cmdr2:
Thanks!

PR #462, opened 2022-11-11
prompt textbox - only vertical resize

JeLuF:
If the resize handle of the textbox is pulled under the result pane, the result pane will obscure that part of the textbox. The handle becomes unreachable and the user can't resize the textbox any more. 

https://discord.com/channels/1014774730907209781/1040411379691176026/1040414658554765402

In older versions of the UI, it was possible to make the textbox wider, and the result pane would become thinner:
image

Since this is no longer the case, horizontal resizing should be disabled.

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -28,6 +28,7 @@ label {
     font-size: 13px;
     margin-bottom: 6px;
     margin-top: 5px;
+    resize: vertical;
 }
 .image_preview_container {
     margin-top: 10pt;


wywhynot:
As the OP in the discord, I want to clarify that I DO want horizontal resizing, just let there be a limit to how much

cmdr2:
Thanks to @Bilbo's Last Clean Doily , we've just pushed a fix to beta. Seems to be working for me now. Closing this PR, since we've got back the old resizing behavior (vertical and horizontal)

PR #463, opened 2022-11-11
Adding Use Settings

patriceac:
Adds support for Use Settings. Restores task configuration, including modifiers, source image, and inpainting mask.


cmdr2:
Thanks! :)

PR #474, opened 2022-11-14
Add paste button next to copy button

JeLuF:
Adds a paste button next to the copy to clipboard button on top of the image settings.
The button accepts either JSON content or autosave sidecar text file content - the same
content that drag and drop accepts as input.


cmdr2:
Thanks!

RenaKunisaki:
Am I missing something? I'm on the beta channel but I don't see this button. Drag and drop shows a + cursor but doesn't do anything.

PR #475, opened 2022-11-14
🔥Fix system info for CPU mode

JeLuF:
System info crashes the UI if the render device is not a GPU.
CPU info doesn't contain mem_free => toFixed fails => exception raised => catch() sets server state to offline.


cmdr2:
Thanks, sorry about that :-)

PR #476, opened 2022-11-14
Incr. Server State Validtiy to 90s

JeLuF:
By default, healthCheck() is run every 5s. On background tabs, this may get extended. My tests have shown pings every 60s. The ping was older than 10s, so the condition in line 490 evaluates to `false` and the client tries to access the stream before the server is ready. By increasing the validity this can be avoided - at least until the browser runs the healthcheck even less often.

See https://discord.com/channels/1014774730907209781/1041811939380178964/1041812021018120262 for the analysis.

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -444,7 +444,7 @@ async function doMakeImage(task) {
     const RETRY_DELAY_IF_BUFFER_IS_EMPTY = 1000 // ms
     const RETRY_DELAY_IF_SERVER_IS_BUSY = 30 * 1000 // ms, status_code 503, already a task running
     const TASK_START_DELAY_ON_SERVER = 1500 // ms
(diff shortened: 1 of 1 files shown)


madrang:
Long tasks mitigation on desktop and mobile are so different from browser to browser.
iDevices will kill the function, Andoid slows it to a crawl, Linux will freeze it fine but on reload will redo the init again.
At least setting the timeout delay to 90 seconds will make it happen less on desktop but wont do nothing for mobile devices where the deepsleep could last many minutes before it's awaken again.

In short, you don't want to trigger long tasks mitigation by returning execution back to the browser very often.
But I did raise the SERVER_STATE_VALIDITY_DURATION inside engine.js too just in case.
I don't want to ever see that bug again.

cmdr2:
Thanks!

PR #478, opened 2022-11-15
Changed update_render_threads to use SetAppConfigRequest.

madrang:
Changed update_render_threads to use SetAppConfigRequest to set which devices are active.
Keep ImageRequest.render_device for affinity only. (Send a task to an already active device.)


cmdr2:
Thanks!

PR #480, opened 2022-11-15
Handle %TMP% and %TEMP%

JeLuF:
`conda activate` fails if %TMP% or %TEMP% contains spaces. So instead of setting the temp variables in the installer steps, we need to set it at the beginning of the script. 

Would this break anything? 
https://discord.com/channels/1014774730907209781/1042190192049668106/1042190192049668106

### DIFF
--- a/scripts/on_sd_start.bat
+++ b/scripts/on_sd_start.bat
@@ -1,5 +1,8 @@
 @echo off
 
+@REM Caution, this file will make your eyes and brain bleed. It's such an unholy mess.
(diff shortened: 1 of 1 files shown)


cmdr2:
This should be okay I think. If it works fine on our PCs, then we can push it to production and see if it fails. I don't think it'll affect any existing installations, because this is just the temp folder. But let's push and see.

JeLuF:
Did a few test runs, looks fine.

PR #488, opened 2022-11-16
Add Change history

JeLuF:
Release notes are important to inform users about added functionality. A change history will make it easier for the release coordinator to compile release notes, instead of having to go through all the commits (several hundred for the last release).


cmdr2:
Agreed.

PR #492, opened 2022-11-16
Picklescan of model files

JeLuF:
During getModel(), the server picklescans the model files for potential malicious code in the pickled python objects. If a malicious file is found, the web UI will show a big red error message, the makeImage button will be disabled, and the user must remove the malicious file and reload the UI page.


cmdr2:
Thanks!

PR #494, opened 2022-11-16
Link to the troubleshooting guide if there are problems importing GFPGANer

JeLuF:
image

The `input()` call is needed since uvicorn will print long stack traces that would cover the above error message.

### DIFF
--- a/ui/sd_internal/runtime.py
+++ b/ui/sd_internal/runtime.py
@@ -24,7 +24,27 @@ from ldm.util import instantiate_from_config
 from optimizedSD.optimUtils import split_weighted_subprompts
 from transformers import logging
 
-from gfpgan import GFPGANer
+try:
+    from gfpgan import GFPGANer
+except Exception as e:
+    import rich
+    rich.print(traceback.format_exc())
+    rich.print(str(e))
+    rich.print()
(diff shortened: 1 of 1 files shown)


cmdr2:
Nice work! :)

Two comments:
1. I dug into this problem a bit, and turns out we just need to edit a text file. No need to reinstall those 4 packages. I've updated the troubleshooting guide: https://github.com/cmdr2/stable-diffusion-ui/wiki/Troubleshooting#modulenotfounderror-no-module-named-gfpgan

In short, the `stable-diffusion-ui\stable-diffusion\env\Lib\site-packages\easy-install.pth` file keeps the absolute links to the four files. So just changing them to the new folder location fixes the gfpgan error.

2. I feel we can just auto-fix this problem in the installer. At the end of `on_sd_start` (once the `PYTHONPATH` is set), we could call a simple python program that checks whether the module import fails, and whether the `stable-diffusion\src\gfpgan` folder exists. If so, just edit that file to update the `easy-install.pth` file.

This way, we won't have to write this logic in ugly batch/bash. `on_sd_start` could call this program, say, `fix_if_folder_moved.py` at the end after PYTHONPATH is set.

JeLuF:
Autofix is a better way to address this.

PR #495, opened 2022-11-17
Fix pasting in Firefox.

madrang:
Should not display button until a fallback is added.
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility


cmdr2:
Thanks!

PR #502, opened 2022-11-18
Restore VAE setting

patriceac:


cmdr2:
Thanks, the VAE changes look good. But only the `dnd.js` file is necessary, the rest are getting dragged along unnecessarily.

I've given some suggestions on https://github.com/patriceac/stable-diffusion-ui/pull/1 for how to fix this. You don't need to close this PR.

patriceac:
Created a new PR with just the right changes.

PR #510, opened 2022-11-19
Add network settings to the UI

JeLuF:
- Allow users to choose the uvicorn port
- Allow users to restrict uvicorn to only listen on localhost

image


cmdr2:
Nice, thanks! :)

PR #518, opened 2022-11-20
Fix duplicate custom modifiers activation states

patriceac:
Fixing activation state for custom modifier cards sharing the same tag where only one of the cards gets (de)activated.

### DIFF
--- a/ui/media/js/image-modifiers.js
+++ b/ui/media/js/image-modifiers.js
@@ -90,9 +90,7 @@ function createModifierGroup(modifierGroup, initiallyExpanded) {
                 if (activeTags.map(x => x.name).includes(modifierName)) {
                     // remove modifier from active array
                     activeTags = activeTags.filter(x => x.name != modifierName)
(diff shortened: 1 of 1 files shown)


ui/media/js/image-modifiers.js
```diff
@@ -234,6 +228,22 @@ function refreshTagsList() {
     editorModifierTagsList.appendChild(brk)
 }
 
+function toggleCardState(modifierName, makeActive) {
+    document.querySelector('#editor-modifiers').querySelectorAll('.modifier-card').forEach(card => {
```
cmdr2, line 232: Instead of looping over every item, even after a card has been found, how about:

```
let cards = document.querySelector('#editor-modifiers').querySelectorAll('.modifier-card').filter(card => card.querySelector('.modifier-card-label').innerText === modifierName)
cards.forEach(card => {
  if (makeActive) ...
}

patriceac: Test comment

PR #519, opened 2022-11-20
Rewrite easy_install.pth on each start

JeLuF:
Fixes GFPGANer errors when the env has been moved.


cmdr2:
Thanks!

cmdr2:
Interestingly, Linux isn't as portable. A renamed folder shows `uvicorn not found`. Need to check if we're missing any PATH entries. But this `easy_install.pth` should be applicable on Linux as well.

PR #520, opened 2022-11-20
Fix the gfpgan fix for multi-gpu

madrang:
After this changes, all cards can use Fix Faces, but only one at a time will be able to run the filter.
If two try to run the filter at the same time, one will wait until the other finishes before starting.

### DIFF
--- a/ui/sd_internal/runtime.py
+++ b/ui/sd_internal/runtime.py
@@ -28,6 +28,8 @@ from gfpgan import GFPGANer
 from basicsr.archs.rrdbnet_arch import RRDBNet
 from realesrgan import RealESRGANer
 
+from threading import Lock
+
 import uuid
 
 logging.set_verbosity_error()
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks, this makes sense. I'll merge this in, but it needs a few more lines (I'll add them in a second commit).

Essentially, I applied the hack *before* calling the GFPGANer constructor, because a few other models and variables use the `retinaface.device` field. But if we apply the hack afterwards, like this PR does, then those variables will continue to stay on `cuda`.

I'll fix this in a second commit after merging this PR, by moving those variables directly to the correct device:

```
    thread_data.model_gfpgan.face_helper.device = retinaface.device
    thread_data.model_gfpgan.face_helper.face_det = thread_data.model_gfpgan.face_helper.face_det.to(thread_data.device)
    thread_data.model_gfpgan.face_helper.face_det.mean_tensor = thread_data.model_gfpgan.face_helper.face_det.mean_tensor.to(thread_data.device)
    thread_data.model_gfpgan.face_helper.face_parse = thread_data.model_gfpgan.face_helper.face_parse.to(thread_data.device)```

PR #528, opened 2022-11-22
Add support for custom modifiers to d&d and clipboard

patriceac:
Add support for custom modifiers to d&d and clipboard and remove now-redundant code in restoreTaskToUI.


cmdr2:
Thanks!

PR #531, opened 2022-11-22
Confirm 'Clear All' and 'Stop Task'

JeLuF:
Ask for a confimation before clearing the results pane or stopping a render task. The dialog can be skipped by holding down the shift key while clicking on the button. The confirmation dialog shows a hint that explains this shortcut.


JeLuF:
Sounds reasonable, I'll add a setting.

cmdr2:
Thanks!

PR #534, opened 2022-11-23
Custom modifiers as a plugin

patriceac:
Moves the custom modifiers parsing logic to a plugin and enables user plugin for custom modifiers, e.g. to override or augment existing logic. This code on its own maintains the current behavior.


ui/plugins/ui/custom-modifiers.plugin.js
```diff
@@ -0,0 +1,33 @@
+(function() {
+    PLUGINS['MODIFIERS_LOAD'] = []
```
JeLuF, line 2: This would make it impossible to have other core plugins that provide additional modifiers. An initilization shouldn't be required here, since this already happens in plugins.js:28

patriceac: I see. Didn't realize the array was already being initialized. I thought it was just being declared in some kind of namespace. But I see it now. Thanks!

PR updated.

patriceac: On second thought that logic was actually correct. If this initialization doesn't happen here, the subsequent push() will fail because PLUGINS['MODIFIERS_LOAD'] is undeclared at this point in the flow. However this shouldn't be a concern since core plugins are loaded before user plugins.

patriceac: Let me take that back. That was confirmation bias, my apologies for that. I suspect there's a race condition between the loading of the plugins and the loading of the custom modifiers. Looking into it.

patriceac: Issue fixed. It was on my side (was testing from the wrong branch).
All good now. Apologies for the back and forth.

PR #540, opened 2022-11-24
Set execution bit for on_sd_start.sh

JeLuF:
The x-bit is missing in the new zip file.
 100644 → 100755


cucumberian:
scripts/on_env_start.sh: line 43: ./scripts/on_sd_start.sh: Permission denied

alternatively line 43 maybe like:
```bash ./scripts/on_sd_start.sh```

HPPinata (approved):

cmdr2:
Thanks, I saw this PR a bit too late, and fixed this already. :) I'm still not sure how the exec bit changed in the first place.

PR #544, opened 2022-11-24
Work in progress: HTTP download for server.py

JeLuF:
Supports download of http and https files, resumes downloads automatically. The code works, but is missing features.

Todo:
- pip install aiofile
- remote interface for download status (currently console only)
- UI


JeLuF:
### Todos:
Mandatory:
- Search functionality. Currently hardcoded to "top downloads".
- NSFW filter.
- "Back" button to return from details to results. Browser back button support?
- Saving tokens of new model doesn't work. Probably stuff happens in the wrong order.
- test multiple parallel downloads
- review function names. Some names don't match the functions purpose any more
- catalog button doesn't work any more after download of the first model. Debug console shows that the event listener is bound to the right function.

Optional
- On UI launch, if the list of models is empty, offer to download one of the base models (1.4, 1.5, 2.0)
- Currently, if a download is ongoing, and the user reloads the page, the tokens will not be configured, sort out how to fix. Also, download status will not be shown in the UI any more
- Test multiple concurrent downloads.
- new SVG civitai logo

PR #557, opened 2022-11-26
Added web manifest to allow installing the Url as a web app.

madrang:
Specify `theme_color` as black to work with any theme.
Edit: Now uses the theme colors for the app borders.
image


madrang:
Selection_895
Selection_894
Selection_893

madrang:
image
image

PR #561, opened 2022-11-27
Fix RuntimeError: Expected all tensors to be on the same device, but …

madrang:
Fix RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument index in method wrapper__index_select)

When running on cuda:N, when N is not 0.


cmdr2:
Edited the upstream SD 2 repo to include the register_buffer() fix. Will update the dpm2 solver soon

madrang:
Closed, replaced by https://github.com/cmdr2/stable-diffusion-ui/pull/591

PR #566, opened 2022-11-28
Visual feedback on button click

patriceac:
When there are too many tasks and the top of the list is not visible, there is no visual feedback that a task has been successfully added to the queue.

Adding a subtle visual feedback on buttons upon click to reflect that the mouse event was taken into account.


cmdr2:
Thanks! Very nice

PR #588, opened 2022-12-01
Fix weird scrolling when using a pen

JeLuF:
With a pen, tipping on a browser page, waiting a short moment, and then moving the pen scrolls the page. Call event.preventDefault() to disable this default behaviour for events in the canvas area.

https://discord.com/channels/1014774730907209781/1021695193499582494/1047965838482874438


cmdr2:
Thanks!

PR #596, opened 2022-12-02
Fx: Highlight artist modifiers when clicked

JeLuF:
Artist modifiers, with the exception of Artstation (the first one), don't have the outline when selected. All the other modifiers, above or below, seem to work as intended

https://discord.com/channels/1014774730907209781/1014774732018683927/1048343258775949322


cmdr2:
Thanks!

PR #609, opened 2022-12-04
Fix make image button behavior

patriceac:
renameMakeImageButton code doesn't show "Enqueue Next n Images" when a task starts or is already running because it fails to properly identify a task is running in the first place.

Repro steps:
- Start an image generation task.

Expected result: the make image button reads "Enqueue Next n Images". Actual result: the make image button reads "Make n Images".

- While the task is running, change the number of desired images.

Expected result: the make image button reads "Enqueue Next n Images". Actual result: the make image button reads "Make n Images".

Also, the logic to construct the string is awkwardly English-specific. Making it more localization-friendly (and easier to read).

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -684,7 +684,6 @@ async function checkTasks() {
     setStatus('request', 'fetching..')
 
     stopImageBtn.style.display = 'block'
-    renameMakeImageButton()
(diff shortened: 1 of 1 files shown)


patriceac:
Looks like something has changed and this code is no longer correct. Will take a look at it.

PR #611, opened 2022-12-06
Add events for task creation and task restoration

patriceac:
Add some events to support plugin development.
- task restoration to UI
- new task queuing

Right now this will help with the localized prompt plugin (e.g. translate a non-English prompt to English and run the resulting task).


patriceac:
Looks like something changed and this code is no longer correct. Will have another look.

PR #619, opened 2022-12-06
Hypernetwork support

C0bra5:
Adds support for hypernetworks. Hypernetworks are stored in /models/hypernetworks

I think you have a way to patch the contents of the stable-diffusion folder (patch files in sd_internal), but I have no clue how that works. Therefore I'm just hijacking the CrossAttention.forward function in hypernetwork.py and hope it's good enough.

I'm not really familiar with NN/ML concepts. I wouldn't be surprised if I got something wrong while in the implementation at some point. This is mostly just a port for AUTOMATIC1111's hypernetwork without the training-related code.


cmdr2:
Thanks! :)

PR #627, opened 2022-12-07
allow aarch64

HPPinata:
some systems report "aarch64" on uname -m. Allow them to run this script.

Since $OS_ARCH gets set to "aarch64" on linux systems later if it was "arm64" just use that.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -21,7 +21,8 @@ OS_ARCH=$(uname -m)
 case "${OS_ARCH}" in
     x86_64*)    OS_ARCH="64";;
     arm64*)     OS_ARCH="arm64";;
-    *)          echo "Unknown system architecture: $OS_ARCH! This script runs only on x86_64 or arm64" && exit
+    aarch64*)   OS_ARCH="arm64";;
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks, does this project work with aarch64? Is conda able to find and install the packages correctly?

HPPinata:
I just saw arm64 listed and thought the script would work. Did some testing today in an emulated arm64/aarch64 docker container (openSUSE tumbleweed).
It works up to the stable-diffusion setup but then pytorch 1.11.0 and torchvision 0.12.0 are both not available and just swapping them for 1.12.1 and 0.13.1 did not work. I have no idea what I´m doing, so I don´t think I´ll be able to solve this.
This PR was just supposed to fix what I thought was an oversight (especially considering there is a check for linux+arm64 but not aarch64 in the script, even though they are just different names for the same thing) but I probably should have done some testing first.
If someone who actually knows how this stuff works wants to take a look feel free to do so, otherwise this can be closed (or I can open an enhancement issue for arm support, because it *should* be possible)

cmdr2:
Thanks for trying this out! I'll close this since this PR isn't applicable anymore.

PR #632, opened 2022-12-09
Scrolling panes

patriceac:
Decouple the editor and the preview panes. Scrollbars color updated as well as requested.


cmdr2:
Very nice work! :)

PR #638, opened 2022-12-10
show initimg in task list

JeLuF:
Show a small thumbnail of the init image in the task list
image

Show the full size version of the image on mouse hover
image


JeLuF:
@madrang says I shouldn't use `$()` will rewrite those parts with vanilla JS

cmdr2:
This is great, thanks! :) jQuery usage - I don't really care

PR #641, opened 2022-12-11
Pause button

JeLuF:
image

Pause button that will prevent new render tasks from being started. Running tasks will not be interrupted, they can complete.

https://discord.com/channels/1014774730907209781/1014774732018683928/1050977439179874375
https://discord.com/channels/1014774730907209781/1014774732018683928/1051282461419642991

### DIFF
--- a/ui/media/js/engine.js
+++ b/ui/media/js/engine.js
@@ -437,7 +437,7 @@
          * @memberof Task
          */
         async post(url, timeout=-1) {
-	    if (typeof pauseClient != 'undefined' && pauseClient===true) {
+	    if (typeof pauseClient != 'undefined' && pauseClient===true && typeof resumeClient === 'function') {
 	        await resumeClient()
 	    }
             if(this.status !== TaskStatus.init && this.status !== TaskStatus.pending) {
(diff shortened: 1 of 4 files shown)


cmdr2:
Thanks!

I agree with Patrice about this going next to the "stop all" button. Should also rename to "Pause All", for consistency.

cmdr2:
Nicely done btw

ui/media/js/engine.js
```diff
@@ -437,6 +437,9 @@
          * @memberof Task
          */
         async post(url, timeout=-1) {
+	    if (typeof pauseClient != 'undefined' && pauseClient===true && typeof resumeClient === 'function') {
+	        await resumeClient()
```
madrang, line 441: The event callback onIdle is now async in main.js
https://github.com/cmdr2/stable-diffusion-ui/pull/643

Calling `await resumeClient()` from onIdle will now do just the same and would keep all the UI code inside main.js

JeLuF:
Removed changes from engine.js, moved the pause check into onIdle() as @madrang suggested.
Moved the pause button next to the Stop All button, renamed to Pause All.
image

madrang:
Look great.
I made a small change to reduce that small delay when resuming.
https://github.com/cmdr2/stable-diffusion-ui/pull/649

cmdr2:
Thanks!

PR #644, opened 2022-12-11
Removing jQuery dependency from scrolling panes

patriceac:
TIL that we want to remove any jQuery dependency. This change removes the jQuery dependency introduced by scrolling panes. No change in UX.

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -78,8 +78,8 @@ function updatePreviewSize() {
     }
     lastScrollTop = scrollTop
 
-    $('#editor').css('top', Math.max(-window.pageYOffset + $("#tab-container").offset().top + $('#tab-container').outerHeight(true), 0) + 'px')
-    $('#editor').css('bottom', Math.max($(window).height() - ($("#footer .line-separator").offset().top - $(document).scrollTop()), 0) + 'px')
+    editorContainer.style.top = Math.max(-window.pageYOffset + document.querySelector('#tab-container').offsetTop + document.querySelector('#tab-container').offsetHeight, 0) + 'px'
+    editorContainer.style.bottom =  Math.max(window.innerHeight - (document.querySelector('#footer').getElementsByClassName('line-separator')[0].offsetTop - window.scrollY), 0) + 'px'
 };
 
 let imagePreview = document.querySelector("#preview")


ui/media/js/main.js
```diff
@@ -78,8 +78,8 @@ function updatePreviewSize() {
     }
     lastScrollTop = scrollTop
 
-    $('#editor').css('top', Math.max(-window.pageYOffset + $("#tab-container").offset().top + $('#tab-container').outerHeight(true), 0) + 'px')
-    $('#editor').css('bottom', Math.max($(window).height() - ($("#footer .line-separator").offset().top - $(document).scrollTop()), 0) + 'px')
+    editorContainer.style.top = Math.max(-window.pageYOffset + document.querySelector('#tab-container').offsetTop + document.querySelector('#tab-container').offsetHeight, 0) + 'px'
+    editorContainer.style.bottom =  Math.max(window.innerHeight - (document.querySelector('#footer').getElementsByClassName('line-separator')[0].offsetTop - window.scrollY), 0) + 'px'
```
cmdr2, line 82: I think jquery-like selectors work. Wouldn't this work too?

`document.querySelector('#footer .line-separator').offsetTop`

cmdr2:
Atleast in this repo, I've got no problems with jQuery being around. Others are free to decide for their forks and projects, but it's not a rule or mandate here.

But yeah I agree with this PR, because we can use `document.querySelector()` for most of the things, don't need jQuery for that.

ui/media/js/main.js
patriceac: I don't think it would work, sorry.

document.querySelector('#footer .line-separator') is looking for a single element with both id #footer _and_ class line-separator. I'm looking for an element with class line-separator _within_ the #footer element. Close, but not quite the same thing. :)

cmdr2: Isn't that what was used in the jQuery version? `$("#footer .line-separator")`

cmdr2: Just tried it out, both give the same result. AFAIK, `#footer .line-separator` works just like it works in jQuery, it isn't AND, but selects a nested element as well.

patriceac: Odd... Didn't work in my testing. Anyway, this feature is out so doesn't matter anymore.

patriceac:
As per recent exchange...

PR #649, opened 2022-12-11
Reduce delay when getting out of idle/pause

madrang:
Start the countdown to IDLE_COOLDOWN before idleEventPromise is completed (pause removed)

Faster resuming of tasks when removing pause.


cmdr2:
Thanks

PR #650, opened 2022-12-11
Right click for image buttons

patriceac:
Those image action buttons that show up as soon as the mouse hovers on the image are really intrusive and break basic UX design principles. While I totally see the value of these buttons, the way they are pushed "in my face" feels somewhat obstructive and overdone. :)

No UI/UX change besides the activation on right click.


patriceac:
Taking that one back for now. We should have a setting to allow the user to turn this experience on/off, but settings are currently hardcoded in core.

I can inject my own setting in the UI and store it separately, but wondering if there'd be a better way to integrate plugin settings in general.

PR #662, opened 2022-12-15
Linux uses .zip, not .tar.xz (Fixes #657)

### LINKED ISSUE #657: Documentation for Linux - incorrect
**Describe the bug**
Documentation for the download says that `tar` is used, but the download is in `zip` format.

Minor edit needed.

JeLuF:


cmdr2:
Thanks

PR #669, opened 2022-12-17
CSS only initimg hover, 'use as input' button

JeLuF:
Use a CSS only mouse hover for the initimg in the task bar instead of my previous Javascript based solution (thanks [AssassinJN](https://discord.com/channels/1014774730907209781/1021695193499582494/1053313186234912768))

Add a "Use as Input" button to the initimg hover. Useful for inpainting images which would otherwise be lost.
image


cmdr2:
Thanks!

PR #671, opened 2022-12-18
Allow empty prompts (image modifiers only)

patriceac:
Allows empty prompts as long as there are image modifiers. This allows the user to craft prompts just by using image modifiers if they so wish.


cmdr2:
Thanks!

PR #672, opened 2022-12-18
Human-Readable job times (consistent with approx. time while rendering)

jsuelwald:
Converting "Processed 1 image in 150.62 seconds"
to
"Processed 1 Image in 2 minutes 30 seconds"


jsuelwald:
Will doi that again. Sorry ;)

PR #673, opened 2022-12-18
Change time display on job

jsuelwald:
Change 
> "Processed 1 image in 150.65 seconds"

to

> "Processed 1 Image in 2 minutes 30 seconds" 

to be consistent with the approx. time remaining display while rendering


cmdr2:
Thanks!

PR #675, opened 2022-12-18
Add drag handle

JeLuF:
Add a handle for drag and drop of the taskEntry:
image

Fixes the problems reported in https://discord.com/channels/1014774730907209781/1054076068342005882
Copying any of the text in the task entry became impossible, because trying to mark text causes a dragstart event.


cmdr2:
Thanks!

PR #676, opened 2022-12-18
Return empty list if hostname lookup fails

JeLuF:
When `socket.gethostbyname_ex(socket.gethostname())` fails, the exception causes the entire `/get/system_info` call to fail. Due to this, the UI will not learn about the GPU and thus only CPU rendering will be allowed.

https://discord.com/channels/1014774730907209781/1053805801153773680/1054155074127015956


cmdr2:
Thanks

PR #680, opened 2022-12-19
Drag and Drop Styles

AssassinJN:
Added some functions in main.js and classes in main.css to allow for styling of the drag and drop sort functionality. I made it classed base so that plugins or themes could overwrite the styling as desired.


cmdr2:
Thanks!

PR #685, opened 2022-12-22
Mdiller bugfixes

mdiller:
- fixed theme css properties not being updated properly
- fixed opacity on inpainter to be 100% by default so no weird erasing
- added a fill action (mainly useful for inpainting everything except a certain part, but also nice for setting background when drawing)
- updated inpainter to not auto-clear itself whenever you draw etc

### DIFF
--- a/ui/media/js/image-editor.js
+++ b/ui/media/js/image-editor.js
@@ -100,6 +100,18 @@ const IMAGE_EDITOR_TOOLS = [
 
 const IMAGE_EDITOR_ACTIONS = [
 	{
+		id: "fill",
+		name: "Fill",
+		icon: "fa-solid fa-fill-drip",
+		handler: (editor) => {
+			editor.ctx_current.globalCompositeOperation = "source-over"
+			editor.ctx_current.rect(0, 0, editor.width, editor.height)
+			editor.ctx_current.fill()
+			editor.setBrush()
+		},
+		trackHistory: true
+	},
+	{
 		id: "clear",
 		name: "Clear",
 		icon: "fa-solid fa-xmark",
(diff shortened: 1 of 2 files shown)


ui/media/js/image-editor.js
```diff
@@ -99,6 +99,18 @@ const IMAGE_EDITOR_TOOLS = [
...
 
 const IMAGE_EDITOR_ACTIONS = [
+	{
+		id: "fill",
+		name: "Fill",
+		icon: "fa-solid fa-fill-drip",
+		handler: (editor) => {
+			editor.ctx_current.globalCompositeOperation = "source-over"
+			editor.ctx_current.rect(0, 0, editor.width, editor.height)
+			editor.ctx_current.fill()
```
cmdr2, line 109: Won't this fill the entire image, instead of just the pixels bounding a region (like how fill works in other paint tools)?

mdiller: Yeah this is just a new action button like "Clear" instead of a tool like "Draw". Do you think this isn't useful enough to be a separate action, and rather I implement the Fill tool instead? Alternatively I could rename this new action button.

JeLuF: Personally, I'd like to see an option to fill the entire image with a non-white color. I've been painting backgrounds more than once. I think this button should be kept, but renamed. "Clear (with color)", "Fill all", "Paint all", using fa-paint-roller as icon?

mdiller: That sounds good to me! I can update it with that icon and that text. and then it sounds like the fill tool is also wanted too, so I could add that to this PR. @cmdr2 Sound good?

cmdr2: Yeah

mdiller: @cmdr2 by the way I committed some changes as requested. The algorithm for the fill tool is the one you linked in discord. took a while to get working but seems to work well now. it had some issues with getting itself into an infinite loop but I think I've fixed that now.

cmdr2: Thanks! Taking a look now, sorry was out on my winter break.

cmdr2:
Thanks! :) Testing it out on a branch, and will merge to beta after that.

PR #691, opened 2022-12-23
Avoid guidance scale "1.0"

JeLuF:
Using a guidance scale of 1.0 will cause an exception in the renderer and return a very confusing error message. https://discord.com/channels/1014774730907209781/1028195513377509376


cmdr2:
Thanks :)

PR #692, opened 2022-12-24
Add "Remove" button to each image's hover menu (Fixes #682)

### LINKED ISSUE #682: Add button to discard or hide individual images
**Is your feature request related to a problem? Please describe.**

When you generate a lot of images, it would help a lot if you can filter out undesired results as you spot them. 

**Describe the solution you'd like**

I would like to see a "Discard" button on every image, which will remove or hide it from the results. The button should be usable while new images are still being generated for the same query.

JeLuF:
When generating larger batches, I often want to remove the bad results in order to concentrate on the good ones.


cmdr2:
Thanks!

PR #702, opened 2022-12-26
Fix copy image settings to clipboard

patriceac:
Fixing the copy settings to clipboard feature. Regression was caused by the processing of the legacy turbo field, which I understand to now be obsolete.


cmdr2:
Thanks

PR #707, opened 2022-12-26
Adding image modifiers events

patriceac:
Adding events to allow plugins to listen for image modifiers loaded and refreshed events respectively.


cmdr2:
Thanks!

PR #715, opened 2022-12-28
Convert [ to \[ so the logging backend...

jsuelwald:
... doesn't interpret that as a colour or other command

### DIFF
--- a/ui/easydiffusion/renderer.py
+++ b/ui/easydiffusion/renderer.py
@@ -29,8 +29,8 @@ def init(device):
 
 def make_images(req: GenerateImageRequest, task_data: TaskData, data_queue: queue.Queue, task_temp_images: list, step_callback):
     context.stop_processing = False
-    log.info(f'request: {pprint.pformat(get_printable_request(req))}')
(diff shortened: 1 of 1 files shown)


ui/easydiffusion/renderer.py
```diff
@@ -29,8 +29,8 @@ def init(device):
 
 def make_images(req: GenerateImageRequest, task_data: TaskData, data_queue: queue.Queue, task_temp_images: list, step_callback):
     context.stop_processing = False
-    log.info(f'request: {pprint.pformat(get_printable_request(req))}')
-    log.info(f'task data: {pprint.pformat(task_data.dict())}')
+    log.info(f'request: {pprint.pformat(get_printable_request(req).replace("[","\["))}')
+    log.info(f'task data: {pprint.pformat(task_data.dict().replace("[","\["))}')
```
cmdr2, line 33: Looks like a typo: `task_data.dict().replace`

That's calling replace on the dict. I think you're missing a closing ) before it.

cmdr2:
Thanks, merged!

PR #717, opened 2022-12-28
Restore download link for Linux in beta, ...

jsuelwald:
and make shellscripts in scripts/ executable


JeLuF:
The chmod is not needed when the zip file is unpacked using `unzip`, but some obscure tools (`midnight-commander`) apparently ignore the mode settings in zip files.

cmdr2:
Thanks!

PR #719, opened 2022-12-28
Fix for duplicate images

patriceac:
When eye correction, upscaling, and only show filtered image are ALL disabled, the UI still generates two of the same image, and increments the second's seed by 1 (although it's the same image). It doesn't perform an additional process, but the item is shown twice.


JeLuF:
This fixes this bug: https://discord.com/channels/1014774730907209781/1021695193499582494/1057733174681141248

cmdr2:
Thanks!

PR #720, opened 2022-12-28
Proper restoration of inactive image modifiers

patriceac:
Inactive image modifiers (right click on image tag) are not properly restored by Use Settings and Copy/Paste settings. This PR fixes that.

image


cmdr2:
Thanks!

PR #724, opened 2022-12-29
Img2img settings restoration

patriceac:
Fix source image, mask, and color profile restoration for use settings, copy/paste, and d&d.

Addresses inpainting issue reported here and then some: https://discord.com/channels/1014774730907209781/1014774732018683928/1055428191021432872

### DIFF
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -304,11 +304,7 @@ function restoreTaskToUI(task, fieldsToSkip) {
     // restore the original tag
     promptField.value = task.reqBody.original_prompt || task.reqBody.prompt
 
-    // properly reset missing fields
-    if (!('use_hypernetwork_model' in task.reqBody)) {
-        hypernetworkModelField.value = ""
-        hypernetworkModelField.dispatchEvent(new Event("change"))
-    }
+    // properly reset checkboxes
     if (!('use_face_correction' in task.reqBody)) {
         useFaceCorrectionField.checked = false
     }
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -326,13 +326,12 @@ function restoreTaskToUI(task, fieldsToSkip) {
     }
     else if (task.reqBody.init_image !== undefined) {
         // listen for inpainter loading event, which happens AFTER the main image loads (which reloads the inpainter)
-        document.addEventListener('imagePainterLoad', function() {
+        initImagePreview.addEventListener('load', function() {
             if (Boolean(task.reqBody.mask)) {
                 imageInpainter.setImg(task.reqBody.mask)
             }
         }, { once: true })
         initImagePreview.src = task.reqBody.init_image
-        initImagePreview.dispatchEvent(new Event("load"))
     }
 }
 function readUI() {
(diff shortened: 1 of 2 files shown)


ui/media/js/dnd.js
```diff
@@ -306,19 +313,25 @@ function restoreTaskToUI(task, fieldsToSkip) {
     }
     if (!('mask' in task.reqBody)) {
         maskSetting.checked = false
+        maskSetting.dispatchEvent(new Event("click"))
```
cmdr2, line 316: Shouldn't this "click" event only be dispatched if maskSetting.checked was true? Otherwise this isn't a state change.

ui/media/js/dnd.js
```diff
@@ -306,19 +313,25 @@ function restoreTaskToUI(task, fieldsToSkip) {
...
+        initImagePreview.dispatchEvent(new Event("load"))
         if (Boolean(task.reqBody.mask)) {
             setTimeout(() => { // add a delay to insure this happens AFTER the main image loads (which reloads the inpainter)
                 imageInpainter.setImg(task.reqBody.mask)
             }, 250)
         }
     }
+    else
+    {
+        initImageClearBtn.dispatchEvent(new Event("click"))
```
cmdr2, line 333: Shouldn't this "click" event only be dispatched if the image contained something (before we set it to empty in line 322)? Otherwise this isn't a state change.

Maybe by checking the regex before we set the image src (i.e. before line 322) to determine whether the image was empty before we set something.

ui/media/js/dnd.js
patriceac: Fixed.

ui/media/js/dnd.js
patriceac: Fixed, and also replaced the hardcoded delay with a smoother and more robust event-driven flow.

ui/media/js/dnd.js
```diff
@@ -304,20 +312,27 @@ function restoreTaskToUI(task, fieldsToSkip) {
...
-        if (Boolean(task.reqBody.mask)) {
-            setTimeout(() => { // add a delay to insure this happens AFTER the main image loads (which reloads the inpainter)
+    // hide/show source picture as needed
+    if (IMAGE_REGEX.test(initImagePreview.src) && task.reqBody.init_image == undefined) {
+        // hide source image
+        initImageClearBtn.dispatchEvent(new Event("click"))
+    }
+    else if (task.reqBody.init_image !== undefined) {
+        // listen for inpainter loading event, which happens AFTER the main image loads (which reloads the inpainter)
+        document.addEventListener('imagePainterLoad', function() {
```
cmdr2, line 329: Why not add a regular 'load' event listener? Like `initImagePreview.addEventListener('load', function..)`.

addEventListener() allows multiple functions to be enqueued for an event (that's why it's addEventListener, not setEventListener).

This way, we won't have to do a custom event dispatch in img2imgLoad().

Also, I don't think we need a custom 'load' dispatch in line 335. A 'load' event should be called automatically when the src image is loaded.

cmdr2:
Nice work, and thanks for your patience! :)

PR #725, opened 2022-12-29
Limit the size of zoomed-in source images

patriceac:
If the source image has a high enough resolution it won't fit on the screen when hovering over it. This simple fix limits the max size so the user always has a chance to see the full image.

image


cmdr2:
Thanks!

PR #727, opened 2022-12-29
Restore the original prompt if provided

patriceac:
Restore the original prompt if provided... including if it's empty now that empty prompts are allowed if there are modifiers. As shown in the below picture, I created a modifier-only prompt, but upon restoring the task the prompt got copied back in the prompt field, which will cause a duplication of strings in the next task (prompt + image modifiers are the same and will be concatenated).

image


cmdr2:
Thanks!

PR #733, opened 2022-12-30
Backend side merge API

JeLuF:
Adds an API endpoint that accepts merge requests.
Adds dot, dash and underscore to the list of allowed characters in filenames.

### DIFF
--- a/ui/easydiffusion/server.py
+++ b/ui/easydiffusion/server.py
@@ -13,7 +13,7 @@ from starlette.responses import FileResponse, JSONResponse, StreamingResponse
 from pydantic import BaseModel
 
 from easydiffusion import app, model_manager, task_manager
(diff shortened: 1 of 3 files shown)


ui/easydiffusion/server.py
```diff
@@ -181,6 +186,25 @@ def render_internal(req: dict):
         log.error(traceback.format_exc())
         raise HTTPException(status_code=500, detail=str(e))
 
+def model_merge_internal(req: dict):
+    try:
+        from sdkit.train.merge_models import merge_models
```
cmdr2, line 191: It should be enough to do `from sdkit.train import merge_models`. It'll import the function, instead of the submodule. This is the recommended way to import (declared in the API), since the inner modules are hidden (implementation details) and can change.

cmdr2:
Thanks!

PR #736, opened 2022-12-30
Enforce an autosave directory

### LINKED ISSUE #597: Server-side settings for the save folder
**Is your feature request related to a problem? Please describe.**
On a multi-user-server, it's problematic that the remote users can define a save folder. They can write to any folder on the server that server.py has access to

**Describe the solution you'd like**
config.bat setting that overwrites the user provided save path. If set, the option in the UI is disabled.

**Additional context**
https://discord.com/channels/1014774730907209781/1014774732018683928/1048353853466222602

> Aytee — today,  22:43 
> Just wanted to briefly bump my minor suggestion that I posted before the holiday weekend but got buried since then:
> With my access from multiple devices, I'd love for the ability to set the Auto-Save Images setting/location from the server-side, rather than it being a per-browser setting. It's sometimes hard for me to remember the full file path I use when I'm on my phone or using another computer.
> Even if I could just hard-code the Auto-Save location in one of the launcher batch file or something (like the IP/port settings), that would be super super helpful for me personally.
> I guess there must be a default location path set somewhere, since it always seems to default to "%userprofile%\Stable Diffusion UI", although I can't seem to find it. 
> But the other half of my problem is if I forget to turn on the actual Auto-save toggle too, I queue up a huge batch without it, and then I have to go individually save the images after-the-fact. This is especially annoying if I queued it on a mobile device.

JeLuF:
Add a config.bat/sh setting `FORCE_SAVE_PATH` that can be used by server admins to restrict auto save to a specific directory. Also useful for users who use different end devices and want to centrally configure the auto save option.
If `FORCE_SAVE_PATH` is set, the auto save options in the UI are disabled.

Fixes #597 
Fixes https://discord.com/channels/1014774730907209781/1052691036981428255

### DIFF
--- a/ui/easydiffusion/app.py
+++ b/ui/easydiffusion/app.py
@@ -4,6 +4,7 @@ import sys
 import json
 import traceback
 import logging
+import shlex
 from rich.logging import RichHandler
 
 from sdkit.utils import log as sdkit_log # hack, so we can overwrite the log config
@@ -36,6 +37,7 @@ CORE_UI_PLUGINS_DIR = os.path.abspath(os.path.join(SD_UI_DIR, 'plugins', 'ui'))
 UI_PLUGINS_SOURCES = ((CORE_UI_PLUGINS_DIR, 'core'), (USER_UI_PLUGINS_DIR, 'user'))
 
 OUTPUT_DIRNAME = "Stable Diffusion UI" # in the user's home folder
+PRESERVE_CONFIG_VARS = ['FORCE_SAVE_PATH', 'FORCE_FULL_PRECISION']
 TASK_TTL = 15 * 60 # Discard last session's task timeout
 APP_CONFIG_DEFAULTS = {
     # auto: selects the cuda device with the most free memory, cuda: use the currently active cuda device.
@@ -88,6 +90,11 @@ def setConfig(config):
         bind_ip = '0.0.0.0' if config['net']['listen_to_network'] else '127.0.0.1'
         config_bat.append(f"@set SD_UI_BIND_IP={bind_ip}")
 
+        # Preserve these variables if they are set
+        for var in PRESERVE_CONFIG_VARS:
+            if os.getenv(var) is not None:
+                config_bat.append(f"@set {var}={os.getenv(var)}")
+
         if len(config_bat) > 0:
             with open(config_bat_path, 'w', encoding='utf-8') as f:
                 f.write('\r\n'.join(config_bat))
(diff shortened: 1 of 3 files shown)


ui/easydiffusion/app.py
```diff
@@ -36,6 +37,7 @@
 UI_PLUGINS_SOURCES = ((CORE_UI_PLUGINS_DIR, 'core'), (USER_UI_PLUGINS_DIR, 'user'))
 
 OUTPUT_DIRNAME = "Stable Diffusion UI" # in the user's home folder
+PRESERVE_CONFIG_VARS = ['FORCE_SAVE_PATH', 'FORCE_FULL_PRECISION']
```
cmdr2, line 40: Thanks for the bugfix to preserve `FORCE_FULL_PRECISION`!

However I'd like to get rid of writing to config.bat/sh soon, and would prefer `FORCE_SAVE_PATH` to be in `config.json`. Rethinking my own decision, maybe `FORCE_FULL_PRECISION` should also be in `config.json`.

Both of these are server configs. We only chose to not expose to them to users in the UI (to avoid shooting themselves in the foot). But they are still server configs, hence make sense to be in config.json.

config.bat/sh still make sense to be around, as a convenient way for users to set environment variables. But I'd like to get rid of us writing to them, or relying on them. Thanks!

cmdr2:
Thanks, this looks good! Only minor comment - please use config.json (instead of config.bat/sh) to store the autosave dir path.

ui/easydiffusion/app.py
JeLuF: I fully agree that the duplication of config.bat and config.json is not very nice.

My idea behind this PR was: config.bat for server side (read: admin) settings, config.json for everything accessible via the UI, since JSON files aren't very user friendly to be edited, especially since we don't pretty print the JSON string. It's easier to say "Add this line to that file" than to explain the syntax of a JSON file.

FORCE_SAVE_PATH is an admin setting, so it can't be accessible through the UI.

While writing this down, I got an idea: Could add a small command line tool to manipulate the config.json file:
```
E A S Y  D I F F U S I O N
==========================
1. Reset default model
2. Configure auto save path
3. Configure Full Precision

0. Quit

Please choose from the above options:  _
```

JeLuF: We only use variables from config.bat after we've set up and activated our python env. So it should be possible to have a small script `get_config.py`:
```py
import sys
import json

# Load the JSON file
with open(sys.argv[1], 'r') as f:
    data = json.load(f)

# Iterate through the keys and get the corresponding value
value = data
for key in sys.argv[2:]:
    value = value[key]

print(value)
```
And to use it to extract variables from the JSON file for those cases where we really need them in the shell scripts:
```cmd
for /f "delims=" %i in ('python scripts\get_config.py scripts\config.json net listen_port') do set PORT=%i
```

(Todo: default value handling if there's no such key in the config.json file or the file doesn't exist)

PR #742, opened 2023-01-04
Prevent flooding the log with warnings for GPU<3GB

JeLuF:
Show the warning that GPUs with <3GB aren't supported only once instead of every 5s

Fixes https://discord.com/channels/1014774730907209781/1059869326313799700


cmdr2:
Thanks!

PR #744, opened 2023-01-04
return taskEntry.id on createTask

AssassinJN:
I would like to have createTask return the taskEntry.id in order to allow for watchers or callbacks to be able to reference tasks by id more easily.


cmdr2:
Merged, but could you please give an example where the caller can't just get the `task.id` from the `task` object, after calling `createTask(task)`? It's possible I'm forgetting some scenarios. Thanks

PR #745, opened 2023-01-04
Synchronize .img and .txt autosave file names

JeLuF:
Fixes (3) of this list of bugs: https://discord.com/channels/1014774730907209781/1021695193499582494/1060107021870956624

If the autosave files were generated in different seconds, they have non-matching file names in the base64 encoded timestamps. In order to allow automatic processing, these names should match.


cmdr2:
Thanks!

PR #752, opened 2023-01-07
Fix parsing of text file tasks

patriceac:
parseContent(text) doesn't check the text content being passed actually described a task, which causes some corner case scenarios to break (image task settings are incorrectly cleared because an empty image task is created).


cmdr2:
Merged, thanks! I'd have used `.includes()` to get a true/false value for whether a string is contained inside another string.

PR #762, opened 2023-01-11
Fix the restoring of the previous nested model

patriceac:
Nested models are not properly restored upon page reloading.


cmdr2:
Thanks!

PR #767, opened 2023-01-13
Slider for preview image size

JeLuF:
Add a slider to the system settings so that users can configure the max size of thumbnails.
image

image


JeLuF:
image
Powerpoint and similar tools have a zoom slider in the bottom right. 
Tristona wasn't happy about the 70vh yesterday, https://discord.com/channels/1014774730907209781/1021695193499582494/1073340727716749392, she prefers to have the images at their native resolution, even if that's wider/higher than the screen. I'd thus suggest to allow the slider to go over 100vh.

JeLuF:
Moved slider to generate tab:
image

Alternative idea: Third collapsible section "Display Settings" under "Image Settings" and "Image Modifiers". Could also have a toggle to enable/disable the fixed panes PR from Patrice

JeLuF:
Doesn't work on Firefox

JeLuF:
Fixed the problem on Firefox

patriceac:
FWIW, I like having the slider at the bottom-right, because at the top-right it would overlap with the 'Remove' button and/or make this part of the screen feel overcrowded.

cmdr2:
Sorry for the delay. IMO this doesn't need to be visible all the time. A button called "Thumbnail Size" that opens the slider, next to the `Scroll to generated image` option.

The general thinking is that the top-right corner contains "View Options" for the tasks. E.g. `Scroll to generated image`, and `Thumbnail Size`, and (maybe in the future, `Sort By`).

I didn't want to pile extra work on you, so I'm okay with merging this as it is right now, and taking up this task myself later.

cmdr2:
Similar to how most File Managers show the "Icon Size" as a dropdown, rather than a slider that's always visible.

Because in our case, changing the size of the thumbnail is a low-frequency task IMO, so it doesn't need to occupy such a prominent screen-space all the time.

JeLuF:
image

Hover menu. Todo:
- Find a better label for the button. "Display Settings" or "View Options" sounds like a button to view the options.
- Move inline css to stylesheet.
- Better visual separation of the two elements in the hover dialog. 
- fix merge conflict

JeLuF:
Merge conflict fixed, cleanup still needed, but if you want to pick it up from here, feel free. Otherwise I'll update the PR later today.

JeLuF:
Added my idea of CSS. Feel free to edit it. Also look at #923 which uses many of the same CSS classes.

cmdr2:
Thanks!

PR #769, opened 2023-01-13
Fix number on the "Make X images" button

JeLuF:
With this change, the number of prompt variants is taken into account when computing the number of images that will be generated. 
`X = getPrompts().length * numOutputsTotalField.value`

Fixes https://discord.com/channels/1014774730907209781/1059237919178637484
https://discord.com/channels/1014774730907209781/1021695193499582494/1063402892058431518

image

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -702,12 +702,18 @@ async function onTaskStart(task) {
     if (task.batchCount > 1) {
         // Each output render batch needs it's own task reqBody instance to avoid altering the other runs after they are completed.
         newTaskReqBody = Object.assign({}, task.reqBody)
+        if (task.batchesDone == task.batchCount-1) { 
+            // Last batch of the task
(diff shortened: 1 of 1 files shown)


patriceac:
The formula used here is incorrect and underestimates the number of images that will be produced in some scenarios.
Repro steps:
Create a single-prompt task (e.g. 'an astronaut riding a horse') with number of images set to 6 total, 5 in parallel

Expected result: Make image button reads 'Make 10 Images'
Actual result: Make image button reads 'Make 6 Images'

When task is ran, 10 images are generated.

JeLuF:
Fixed the issue reported by @patriceac:
Create a single-prompt task (e.g. 'an astronaut riding a horse') with number of images set to 6 total, 5 in parallel

Previous result: "Make image" creates 10 images
Fixed result: "Make image" creates 6 images as requested (First job will create 5 images in parallel, the second job will only create 1 image)

cmdr2:
Thanks!

PR #771, opened 2023-01-15
Fix restoration of models with subfolders

patriceac:
In dnd.js, when models are restored in the UI, there is code that strips the path from the model file name. Now that we allow models to be hosted in subfolders, this code break the task restoration (e.g. use settings, D&D, copy/paste) because "/my models/model.ckpt" becomes "model.ckpt", which won't be found.

The fix here essentially looks for the presence of '/models/stable-diffusion/', which we do know would exist in early txt files to decide whether to strip the path.

https://discord.com/channels/1014774730907209781/1014780368890630164/1063911477796417557

### DIFF
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -363,14 +363,15 @@ function readUI() {
 }
 function getModelPath(filename, extensions)
 {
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks

PR #775, opened 2023-01-15
When reloading the main model, reaload VAE and hypernetwork

madrang:


cmdr2:
Thanks for the fix! I felt this was actually a bug in sdkit, so this has been fixed in these two commits:
1. https://github.com/easydiffusion/sdkit/commit/05fb4d3a5eb5512c57cc46abe22f8d909ecd4708
2. Fix for the .vae.pt file scan (in ED): https://github.com/cmdr2/stable-diffusion-ui/commit/5089ac5ad102b6d21edcef7205175a933014a80c

PR #779, opened 2023-01-17
Fix card names and toggling

patriceac:
Fix names and toggling for cards starting with "By ", e.g. "By the ocean".

Before: "by " missing from "by the pool", disabled card included in prompt
image

After: "by the pool" displayed as expected in the card, disabled card not included in the prompt
image


patriceac:
Just to confirm, actual Artist names still how as expected:
image

cmdr2:
Thanks!

PR #784, opened 2023-01-17
Warn when running installer from git checkout

JeLuF:
There are a handful of users every week that try to install from a git checkout, see
https://discord.com/channels/1014774730907209781/1065014040058933319
https://discord.com/channels/1014774730907209781/1064148631516614727
https://discord.com/channels/1014774730907209781/1062636187191361666

### DIFF
--- a/scripts/Start Stable Diffusion UI.cmd
+++ b/scripts/Start Stable Diffusion UI.cmd
@@ -18,10 +18,8 @@ if exist "on_sd_start.bat" (
     echo.
     echo ================================================================================
     echo.
-    echo Not Recommended: If you're sure that you want to run the installer from this
-    echo directory, please press any key to continue.
-    echo.
     pause
+    exit /b
 ) 
 
 @rem set legacy installer's PATH, if it exists
--- a/scripts/Start Stable Diffusion UI.cmd
+++ b/scripts/Start Stable Diffusion UI.cmd
@@ -1,7 +1,28 @@
 @echo off
 
 cd /d %~dp0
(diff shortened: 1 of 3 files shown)


scripts/Start Stable Diffusion UI.cmd
```diff
@@ -1,8 +1,29 @@
...
+    echo.
+    echo It looks like you're trying to run the installation script from a source code 
+    echo download. This will not work.
+    echo.
+    echo Recommended: Please close this window and download the installer from
+    echo https://stable-diffusion-ui.github.io/docs/installation/
+    echo.
+    echo ================================================================================
+    echo.
+    echo Not Recommended: If you're sure that you want to run the installer from this
```
cmdr2, line 21: This scenario shouldn't allow the user to continue. Because we really can't run the installer from inside the scripts folder (after checking out the repo). So I'd suggest exiting after letting the user press enter.

rbertus2000: Then maybe it's better to check if the .git directory exists? That way you would know that it's a git checkout.

JeLuF: There are two scenarios. People using `git clone` and then there are people who download the zip file, which doesn't have the .git directory.
image

rbertus2000: yeah, you are right, you can also download the zip... haven't thought about that.

cmdr2:
Thanks!

PR #786, opened 2023-01-18
Searchable models

patriceac:
Makes the models dropdowns searchable (folder or file) and adds a reload models button.
Also makes it easy to add new model dropdowns - e.g. for the model merger - and takes care of refreshing these seamlessly upon reloading.

Something tells me this is going to be a long code review. 😇

image


ui/media/js/searchable-models.js
```diff
@@ -0,0 +1,586 @@
...
+            if (sibling.style.display == 'list-item') return sibling
+            lastSibling = sibling
+            sibling = sibling.nextElementSibling
+        }
+        
+        // no more siblings, look for next parent if any
+        if (lastSibling.nextElementSibling) {
+            return this.findNextSibling(lastSibling.nextElementSibling.firstElementChild.firstElementChild, false)
+        }
+        else if (lastSibling.parentElement.parentElement.nextElementSibling && lastSibling.parentElement.parentElement.nextElementSibling.firstElementChild && lastSibling.parentElement.parentElement.nextElementSibling.firstElementChild.firstElementChild) {
```
cmdr2, line 144: Any way we can avoid this? :)

cmdr2:
Thanks, I'm okay with this PR getting merged, except that the sibling-related function looks really strange and fragile :) I expect you also feel that way (regarding it being a bit fragile/odd), so can we figure out a better way to do the sibling-check?

Otherwise, this one looks good to merge. Thanks

patriceac:
I hated writing this clause tbh. While it's reliable, it's NOT elegant or
easy to read by any stretch of the imagination. Let me think about it more
and see if there's a better way.

ui/media/js/searchable-models.js
patriceac: Rewrote the findSibling() functions for improved readability and with proper comments. Hope you like it. :)

patriceac:
@cmdr2 findSibling() functions updated and conflicts resolved. Please let me know what you think of the updated functions.

cmdr2:
Thanks! :)

PR #789, opened 2023-01-19
Support multiple GFPGAN models

### LINKED ISSUE #265: GFPGAN 1.4 model?
I notice on the https://github.com/TencentARC/GFPGAN
They have a new 1.4 gfpgan model available. The cmdr2 implementation currently includes 1.3 model

Can this updated model be included in the distribution (or is it possible to update the pth file simply by replacing the existing one)?

JeLuF:
Add scanning for models and a dropdown to choose different models from.
image

In some situations, the 1.3 version produces better results, in others, I like the 1.4 version more. I think supporting both makes sense.

This PR might conflict with the changes in #786

Fixes #265


cmdr2:
Thanks!

PR #793, opened 2023-01-24
Fix the tooltip display over image modifier cards

patriceac:
Before:
image

After:
image

Also opportunistically fixed the indentation of the .simple-tooltip class declaration.


cmdr2:
Thanks

PR #794, opened 2023-01-24
Embed metadata

patriceac:
*** Please merge https://github.com/easydiffusion/sdkit/pull/9 before merging this one. ***

This is the ED client part of metadata embedding. It adds 'embed' and 'none' options to the metadata setting and makes none the default (if never set before) because (1) it feels weird to create metadata files by default and (2) embedding by default could cause be problematic if users don't realize it's happening.

Also fixes the disabling of the dropdown in the settings when Save images to disk is toggled off.

### DIFF
--- a/ui/media/js/parameters.js
+++ b/ui/media/js/parameters.js
@@ -58,8 +58,12 @@ var PARAMETERS = [
         type: ParameterType.select,
         label: "Metadata format",
         note: "will be saved to disk in this format",
-        default: "txt",
+        default: "none",
         options: [
+            {
+                value: "none",
+                label: "none"
+            },
             {
                 value: "txt",
                 label: "txt"
@@ -67,6 +71,10 @@ var PARAMETERS = [
             {
                 value: "json",
                 label: "json"
+            },
+            {
+                value: "embed",
+                label: "embed"
             }
         ],
     },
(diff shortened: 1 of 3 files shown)


ui/media/js/parameters.js
```diff
@@ -58,15 +58,23 @@ var PARAMETERS = [
         type: ParameterType.select,
         label: "Metadata format",
         note: "will be saved to disk in this format",
-        default: "txt",
+        default: "none",
```
cmdr2, line 61: I see what you're saying, but this will be a breaking change for a lot of users (who've never touched this setting, and expect it to keep saving their metadata, on projects that they're working on).

So unfortunately I'd prefer to keep this as 'txt'. Especially because I don't have a strong opinion on whether it's better to save metadata files by default or not.

cmdr2:
Thanks, I've merged the sdkit PR. This one looks good to go too, except that I'd prefer to keep the default metadata format as 'txt' (more in the specific comment).

cmdr2:
Also, I just merged https://github.com/cmdr2/stable-diffusion-ui/pull/789, which warns me that it might conflict with this PR

patriceac:
Ooh, I didn't realize that! I was thinking it would just change for new
installs... Yeah, I'm with you on this one, sorry for the confusion.

Will fix that once I'm home tonight.

PR #800, opened 2023-01-28
Making ellipsis' 3 dots

patriceac:
Call me pedantic, but in any language I can think of an ellipsis is made of 3 dots... :)


cmdr2:
Agreed, but 3 dots waste space (and look a bit ugly on single word fragments), when 2 dots convey the meaning.

PR #803, opened 2023-01-28
Add T600 to list of FP only GPUs

JeLuF:
https://discord.com/channels/1014774730907209781/1068948110304354314

### DIFF
--- a/ui/easydiffusion/device_manager.py
+++ b/ui/easydiffusion/device_manager.py
@@ -125,7 +125,7 @@ def needs_to_force_full_precision(context):
         return True
 
     device_name = context.device_name.lower()
-    return (('nvidia' in device_name or 'geforce' in device_name) and (' 1660' in device_name or ' 1650' in device_name or ' t400' in device_name or ' t550' in device_name or ' t600' in device_name or ' t1200' in device_name)) or ('Quadro T2000' in device_name)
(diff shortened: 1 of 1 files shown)


rbertus2000:
I found a full list of GPUs with this chip online:
https://www.techpowerup.com/gpu-specs/nvidia-tu117.g881
We can wait for the next one, or we just add mx450. mx550 and t1000 now.

JeLuF:
The MXC450 and MX550 have 2GB max, we can ignore them

Looking at the code - isn't the T2000 check always `false`?
```py
 device_name = context.device_name.lower()
 'Quadro T2000' in device_name
 ```
 Should be `quadro`, no?

rbertus2000:
> The MXC450 and MX550 have 2GB max, we can ignore them
> 
> Looking at the code - isn't the T2000 check always `false`?
> 
> ```python
>  device_name = context.device_name.lower()
>  'Quadro T2000' in device_name
> ```
> 
> Should be `quadro`, no?

I think so too, but do we even need 'quadro', when we check with txx in device_name?

JeLuF:
Added t1000, and moved `quadro` in the same list as `nvidia` and `geforce`.
It's odd that the T2000 reports itself as Quadro T2000, while the T1200 is apparently  an NVIDIA T1200 or GEFORCE T1200.

cmdr2:
Merged. Thanks @JeLuF and @rbertus2000 for this :)

PR #811, opened 2023-01-31
Removing the ':' after the tooltip icon

patriceac:
This colon after the tooltip icon just feels out of place.

image


cmdr2:
Thanks!

PR #817, opened 2023-02-03
Fix aarch64 (arm64) verification

fernandoisnaldo:
I tried to run the start.sh script in Asahi Linux on a Mac mini with an Apple CPU and I could see that the first failure is verification of CPU architecture.

The script basically says that my architecture is aarch64 and not arm64, to fix this I added a condition that accepts aarch64 because I understand that aarch64 and arm64 are the same thing.

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -21,6 +21,7 @@ OS_ARCH=$(uname -m)
 case "${OS_ARCH}" in
     x86_64*)    OS_ARCH="64";;
     arm64*)     OS_ARCH="arm64";;
+    aarch64*)     OS_ARCH="arm64";;
     *)          echo "Unknown system architecture: $OS_ARCH! This script runs only on x86_64 or arm64" && exit
 esac
 


JeLuF:
Was the installation successful after this change?

fernandoisnaldo:
> Was the installation successful after this change?

No, because of pytorch and torch vision version.

I was going to continue testing on Asahi Linux, but I saw in the code that there is an intention to run this on macOS as well.

If you need a machine to test and run, I have one available.

cmdr2:
Thanks @fernandoisnaldo :)

cmdr2:
@fernandoisnaldo I look forward to your experiments in getting this to run on Mac. Please feel free to drop by the #development channel on our discord server (https://discord.com/invite/u9yhsFmEkB) ! Would love to chat more.

I think a few changes are needed in the code as well (probably in sdkit [1]), to support the `mps` device. InvokeAI is a good reference for this, to understand the changes related to Mac.

[1] https://github.com/easydiffusion/sdkit

PR #824, opened 2023-02-05
Fix 'Pause All' function

JeLuF:
If 'pause all' is clicked during the last scheduled job, the 'resume all' button gets hidden when the jobs terminates, making it impossible to unpause the engine.
https://discord.com/channels/1014774730907209781/1014780368890630164/1071584183417323602


cmdr2:
Thanks!

PR #829, opened 2023-02-06
add random_seed flag to reqBody

Schorny:
expose if the user requested a random seed or used a fixed seed

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -929,7 +929,7 @@ function getCurrentUserRequest() {
 
         reqBody: {
             seed,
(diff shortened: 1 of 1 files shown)


patriceac:
Leaving a comment to be notified when this gets merged.

ui/media/js/main.js
```diff
@@ -929,6 +929,7 @@ function getCurrentUserRequest() {
 
         reqBody: {
             seed,
+            random_seed: randomSeedField.checked,
```
cmdr2, line 932: Please rename this to `used_random_seed`, and I'm good with this change :)

PR #830, opened 2023-02-07
Sort models alphabetically

ogmaresca:
With EasyDiffusion 2.5 models stopped being sorted alphabetically, which makes it hard to find a specific model when you have a lot of them. I'm adding back sorting, as well as making the sort case insensitive.
I also changed it so that folders appear on top. For child folders the grandchild folders appear last, to prevent the UI from looking like

* **ChildFolder**
* **ChildFolder/GrandchildFolder**
  * Model1
  * Model2

when Model2 belongs to ChildFolder and Model1 belongs GrandchildFolder.

I also updated the UI so that folders that only contain more folders will not appear in the select menus.

With sorting:
EasyDiffusion Sorted Models

### DIFF
--- a/ui/easydiffusion/model_manager.py
+++ b/ui/easydiffusion/model_manager.py
@@ -200,7 +200,7 @@ def getModels():
                 scan=scan_directory(entry.path, suffixes, directoriesFirst=False)
 
                 if len(scan) != 0:
-                    tree.append( (entry.name, scan ) )                       
+                    tree.append( (entry.name, scan ) )
         return tree
 
     def listModels(model_type):
--- a/ui/easydiffusion/model_manager.py
+++ b/ui/easydiffusion/model_manager.py
@@ -179,10 +179,10 @@ def getModels():
         "Raised when picklescan reports a problem with a model"
         pass
 
-    def scan_directory(directory, suffixes):
+    def scan_directory(directory, suffixes, directoriesFirst:bool=True):
         nonlocal models_scanned
         tree = []
-        for entry in sorted(os.scandir(directory), key = lambda entry: (entry.is_file(), entry.name.lower())):
+        for entry in sorted(os.scandir(directory), key = lambda entry: (entry.is_file() == directoriesFirst, entry.name.lower())):
             if entry.is_file():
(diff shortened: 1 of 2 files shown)


ui/easydiffusion/model_manager.py
```diff
@@ -197,9 +197,10 @@ def scan_directory(directory, suffixes):
                 known_models[entry.path] = mtime
                 tree.append(entry.name[:-len(matching_suffix)])
             elif entry.is_dir():
-                scan=scan_directory(entry.path, suffixes) 
+                scan=scan_directory(entry.path, suffixes, directoriesFirst=False)
+
                 if len(scan) != 0:
-                    tree.append( (entry.name, scan ) )
+                    tree.append( (entry.name, scan ) )
```
cmdr2, line 203: Please remove the empty spaces :)

cmdr2:
Hi @ogmaresca Thanks for the PR. I just tried it out, and I'm not sure it's giving the right results. Can you please take a look?

For e.g. on my disk, I have two models inside a folder called 'official/2.0', and another model inside 'official/2.0/depth'.

So this is what it should display (which the current UI shows):
image

This is what the PR is showing (which is not correct):
image

For reference, here is my folder (`official/2.0`):
image

And the `official/2.0/depth` folder:
image

But otherwise, I'm really happy with the PR, and the code change. It's in the right direction, look forward to your reply!

ui/easydiffusion/model_manager.py
ogmaresca: @cmdr2 Done

ogmaresca:
@cmdr2  For the folder display order, did you test commit [e7bf2ee](https://github.com/cmdr2/stable-diffusion-ui/pull/830/commits/e7bf2ee58b163b56b7fd94b4423003891e4bed77) from yesterday? In it I fixed the issue so that child directories appear first only from the root directory, and in child directories the models appear first.

Screenshot_20230209_180444

Screenshot_20230209_180917

cmdr2:
@ogmaresca Apologies, I'll try again.

cmdr2:
Thanks, merged!

cmdr2:
Nice work!

PR #831, opened 2023-02-07
Restore VAE model in metadata files and remove prompt strength in txt2img generations

ogmaresca:
it appears that in EasyDiffusion 2.5 the VAE model parameter in metadata files was accidentally removed. I added the missing label so that VAE models are outputted again.
I also removed the `Prompt Strength: 0.8` parameter in metadata files if you're using txt2img, since it's misleading. And I also removed `Hypernetwork Strength: None` when the hypernetwork model is none, since it's unnecessary.


cmdr2:
Thanks @ogmaresca  :)

PR #843, opened 2023-02-08
Add hook to implement custom prompt preprocessors

JeLuF:
https://discord.com/channels/1014774730907209781/1014774732018683928/1072787224850092083

Adds a hook that can be used to extend the client side prompt preprocessor, before the prompt is passed on to the tasks.

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -1015,6 +1015,8 @@ function getPrompts(prompts) {
     promptsToMake = applyPermuteOperator(promptsToMake)
     promptsToMake = applySetOperator(promptsToMake)
 
+    PLUGINS['GET_PROMPTS_HOOK'].forEach(fn => { promptsToMake = fn(promptsToMake) })
+
     return promptsToMake
 }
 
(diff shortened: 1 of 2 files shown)


patriceac:
Doesn't this exist already? Madrang had previously shared a way to update tasks - including prompts - before task is added to the queue. This is how I did my localized prompt plugin.

Asking just in case- maybe this is a different use case and I'm just missing the point.

Schorny:
id love to have that hook before the applyPermuteOperator call.

the use case is: if you want to implement prompt preprocessing, the current hook of TASK_CREATE comes to late to be reflected in the web gui. you can change the prompt for the backend sd call fine, but any changes you do during TASK_CREATE dont get reflected in the webui.

a hook right before the call to applyPermuteOperator would allow many prompt preprocessing features to be implemented like dynamic prompts or wildcards -> see automatic1111 for examples.

i like the preprocessing to be done BEFORE the ed internal processing, because then you can use the ED internal prompt features in your preprocessing -> imagine a simple js template engine parsing the prompts to allow ifs and such:

https://github.com/adieyal/sd-dynamic-prompts
https://github.com/AUTOMATIC1111/stable-diffusion-webui-wildcards

cmdr2:
Thanks

PR #858, opened 2023-02-11
Remove NSFW Setting from the README

C-Ross:
The NSFW setting in the UI was removed, just removing the statement in the README for clarity.


cmdr2:
It was removed temporarily, and I definitely intend to bring it back. I was experimenting with `diffusers` yesterday, which should help us get this feature back.

Sorry for the confusion, but this setting will be coming back soon,

PR #867, opened 2023-02-12
Linux Installer fixes

JeLuF:
- Copy functions.sh from git repo as well. Otherwise it will use the `main` functions.sh together with the `beta` on_sd_start.sh
- Remove duplicate 'Press any key' call. There is a "Press any key" line in both on_env_start.sh and on_sd_start.sh. Also, use `exec` to replace the bash process.

https://discord.com/channels/1014774730907209781/1074411964966653992


cmdr2:
Thanks!

PR #869, opened 2023-02-13
Model search bug fixes

patriceac:
Bug fixes for model search:
- fix navigation issues with arrow keys when filtering models
- fix the issue with arrow keys jumping several entries after model reloading
- disable autocomplete in model search box
- gfpgan: the list with models doesn't appear under the <input> box
- merge models: As long as no models are selected, the <input> box is very short.
- When searching for models by name, the width of the model list shrinks and is smaller than the <input> element.

https://discord.com/channels/1014774730907209781/1014780368890630164/1074468097135747083


cmdr2:
Thanks!

PR #873, opened 2023-02-13
remove js based sizing

AssassinJN:


AssassinJN:
Remove js based sizing function to be replaced by css in patch-2

PR #880, opened 2023-02-14
Fix functions.sh upgrade, change messages to Easy Diffusion

JeLuF:
`functions.sh` needs to be updated in `on_sd_start.sh` as well, since the update of `on_env_start.sh` happens too late to put the new `functions.sh` file in place.  https://discord.com/channels/1014774730907209781/1021695193499582494/1074959802259755108

Also, replaced all _Stable Diffusion UI_ messages by _Easy Diffusion_ messages, also on Windows.


cmdr2:
Thanks!

PR #881, opened 2023-02-14
Fix reloading of tasks with no file path

patriceac:
In some conditions tasks may be reloaded with an empty file path (e.g. no face correction). This causes the task reloading to throw an exception during the below test because filename is null.

    if (filename.includes('/models/stable-diffusion/')) {
    ...
    }


cmdr2:
Thanks!

PR #884, opened 2023-02-14
show buttons when seed hover

alextraza:
was

became

hover cursor on seed label - show hidden buttons

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -133,7 +133,13 @@ code {
     opacity: 1;
 }
 .imgItemInfo * {
-    margin-bottom: 7px;
+    margin-bottom: 4px;
+}
+.tasksBtns {
+    display: none;
+}
+.imgItemInfo:hover > button {
+    display: inline-block;
 }
 .imgItem .image_clear_btn {
     transform: translate(40%, -50%);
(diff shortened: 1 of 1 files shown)


JeLuF:
How would a user find out about the menu qualities of the seed? A hamburger menu or some other indicator would be needed to make clear that there are more functions available.

alextraza:
add bars
if put fontawesome.css above the main.css, the chevron-down symbol can be used.

fdwr:
@alextraza 🤔🤨 What functionality convenience does the change give the user? I use those buttons over each image *very* frequently, especially "Make Similar Images", "Download", and "Use as input". If this change intends to hide those buttons of an image instance until I first hover over the seed value each and every time for hundreds of times, then I'd reject it for the additional tedium.

cmdr2:
Hi @alextraza Thanks for the PR! I can understand why the suggestion makes sense, but I agree with @fdwr that *in practice* these buttons are high-frequency buttons. I'd like to avoid the extra step of having to click and open a menu each time.

We recently shipped a change to group a lot of related buttons *horizontally*, to reduce the vertical height of the list (especially with plugins). I welcome any suggestion to further improve the design of the buttons, but without adding extra steps for such high-frequency buttons.

Thanks!

PR #885, opened 2023-02-15
Cleaning up event listener that's no longer needed

patriceac:
The event listener instantiates two objects every time the user clicks on the Merge tab. This is no longer needed after AssassinJN's CSS fixes from yesterday.


patriceac:
@JeLuF for awareness

cmdr2:
Thanks!

PR #886, opened 2023-02-15
Add support for disabled state to model dropdown

patriceac:
As per https://discord.com/channels/1014774730907209781/1021695193499582494/1075068193753804831

The only limitation is that we cannot visually gray out the chevron itself because the corresponding font-awesome icon is a Pro icon (https://fontawesome.com/icons/angle-down?s=duotone&f=classic).

### DIFF
--- a/ui/media/css/searchable-models.css
+++ b/ui/media/css/searchable-models.css
@@ -75,6 +75,7 @@
     padding-top: 3px;
     cursor: pointer;
     font-size: 8pt;
+    transition: none;
 }
 
 .model-input {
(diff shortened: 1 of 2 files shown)


JeLuF:
> we cannot visually gray out the chevron itself
Can't you change the text color of the chevron?

patriceac:
Wait- what?!! It works indeed, but why? Aren't those icons images? 👀

patriceac:
Code updated. Thanks JeLuf for pointing out this was possible!

cmdr2:
Thanks!

PR #889, opened 2023-02-15
Make stream_image_progress accept an integer for the rate the progress frames should be generated.

ayunami2000:
Make stream_image_progress accept an integer for the rate the progress frames should be generated.

E.g. The default is 5, but now, it can be streamed more or less often.

This change only affects the API; the UI itself is unaffected.


cmdr2:
Thanks @ayunami2000. I'd prefer adding a different field to `TaskData`, maybe called `stream_image_progress_interval` (int), with a default value of 5.

We can then just use `task_data.stream_image_progress_interval` instead of the hard-coded `5`.

ayunami2000:
Sure, I'll make that change.

ayunami2000:
@cmdr2 Ok, tested and updated.

cmdr2:
Thanks, merged!

PR #893, opened 2023-02-16
Only confirm image deletion once

JeLuF:
The previous code added an event listener per preview image (if live preview is enabled), so that multiple confirmations were required.


cmdr2:
Thanks!

PR #894, opened 2023-02-17
Fix model folders being split up by child folders

ogmaresca:
The new models dropdown sorts a flattened list of models. This can cause issues with subfolders.

Ex: if you have a structure like this:

* foo
  * bar
    *  baz.safetensors
  * aar.safetensors
  * car.safetensors

Then the resulting list would currently look like

* **/foo**
  * aar
* **/foo/bar**
  * baz
* **/foo**
  * car

when it would be better as

* **/foo**
  * aar
  * car
* **/foo/bar**
  * baz

Example of current behavior:
Model Dropdown Current Sort Behavior

New behavior in this PR:
Model Dropdown New Behavior

### DIFF
--- a/ui/media/js/searchable-models.js
+++ b/ui/media/js/searchable-models.js
@@ -597,7 +597,7 @@ class ModelDropdown
                 foldersMap.set(
                     childFolderName,
                     this.createModelNodeList(
-                        `${folderName || '/'}${childFolderName}/`,
+                        `${folderName || ''}/${childFolderName}`,
                         childModels,
                         false,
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks @ogmaresca ! And apologies for the confusion around the dropdown, I appreciate your fixes to the new code.

If you're interested, we'd love to see you on the development channel on our Discord server (https://discord.com/invite/u9yhsFmEkB), since it'll help coordinate with Patrice who's also making changes to the dropdown.

PR #902, opened 2023-02-18
Select but don't empty the search box upon selection

patriceac:
As per your request.

image


cmdr2:
Perfect, good solution :)

PR #904, opened 2023-02-18
🔥 Installer: Fix ESRGAN anime model's path

JeLuF:
The size check fails on every installation. The path name of the check wasn't changed when the model was moved to the models directory.

https://discord.com/channels/1014774730907209781/1076486753394905111


cmdr2:
Thanks, not sure how that happened

cmdr2:
I missed changing that during the models folder refactor

PR #907, opened 2023-02-19
Support WEBP image formats

ogmaresca:
WEBP images are supported by the PIL image library used in sdkit. The format has extremely small file sizes. For example, a 768x760 image that I upscaled 2x to 1536x1536, was

* 189.9 Kb for a 50% quality .webp
* 244.8 Kb for a 50% quality .jpeg
* 637.7 Kb for a 95% quality .webp
* 890.7 Kb for a 95% quality .jpeg
* 3.1 Mb for a .png

All using the same parameters (seed, etc...), just changing the output format. .webp also works for source images in img2img.

I did also test .avif, but PIL throws `ValueError: unknown file extension: .avif`

image

### DIFF
--- a/ui/easydiffusion/types.py
+++ b/ui/easydiffusion/types.py
@@ -38,7 +38,7 @@ class TaskData(BaseModel):
     use_hypernetwork_model: str = None
 
     show_only_filtered_image: bool = False
-    output_format: str = "jpeg"  # or "png"
+    output_format: str = "jpeg"  # or "png" or "webp"
     output_quality: int = 75
     metadata_output_format: str = "txt"  # or "json"
     stream_image_progress: bool = False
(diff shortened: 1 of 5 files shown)


ogmaresca:
In sdkit https://github.com/easydiffusion/sdkit/blob/main/sdkit/utils/image_utils.py#L26, https://github.com/easydiffusion/sdkit/blob/main/sdkit/utils/image_utils.py#L34, and https://github.com/easydiffusion/sdkit/blob/main/sdkit/utils/file_utils.py#L66 may or may need to be updated as well. Not sure what they do, since webp images seem to be completely supported with just easydiffusion changes

cmdr2:
Thanks for the PR!

PR #909, opened 2023-02-19
Bugfix for enforce autosave

rbertus2000:
Little bugfix to show save toggle as enabled and to be able to select metadata output.

### DIFF
--- a/ui/media/js/parameters.js
+++ b/ui/media/js/parameters.js
@@ -421,7 +421,7 @@ async function getSystemInfo() {
         if (res['enforce_output_dir'] !== undefined) {
             force = res['enforce_output_dir']
             saveToDiskField.checked = true
-			metadataOutputFormatField.disabled = false
+            metadataOutputFormatField.disabled = false
             saveToDiskField.disabled = force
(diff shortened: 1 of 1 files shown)


ui/media/js/parameters.js
```diff
@@ -420,6 +420,8 @@ async function getSystemInfo() {
         let force = false
         if (res['enforce_output_dir'] !== undefined) {
             force = res['enforce_output_dir']
+            saveToDiskField.checked = true
```
cmdr2, line 423: I think this should be set to true only if `res['enforce_output_dir']` is true (i.e. the `force` variable is true).

And if `force` is false, we shouldn't modify this field (and let the auto-save logic do its thing).

Basically `res['enforce_output_dir']` can be `false` for other reasons (for e.g. if the user never set `force_save_path` in `config.json`), in which case we shouldn't modify this toggle.

Same comment for the next line.

rbertus2000: ok, I'll check if force == true

cmdr2:
Thanks!

PR #910, opened 2023-02-19
Fix the centering of sharpness brushes

patriceac:
Fixing a visual glitch that becomes visible when a plugin adds borders to the brushes to make them more visible. Visually transparent for the base Easy Diffusion UI.

See this for context; https://discord.com/channels/1014774730907209781/1058857864954904607/1076694770845487155

Before:
image

After:
image


cmdr2:
Thanks!

PR #912, opened 2023-02-19
Support custom modifiers with images

ogmaresca:
This change allows adding custom modifiers with images, instead of just text values. It looks for images in a `modifiers` folder and adds them to the object read from `modifiers.json`.  If an image is directly inside of `modifiers` then it will be added to a Modifiers category. Images in direct subfolders will have the folder names as the category. Otherwise images in sub-sub folders etc... will have the folder path be the category (ex: `modifiers/Foo/Bar` will become `Foo/Bar`). The order of categories defined in `modifiers.json` is preserved, and new categories will be added at the end sorted alphabetically.

If the image ends in ` Portrait`, `.Portrait`, `-Portrait`, or `_Portrait` (case insensitive) it will be considered a portrait image, and if it has Landscape instead it would be a landscape image, and the portrait/landscape suffix will be removed from the modifier name. Otherwise the image will be used as both the portrait and landscape.

I am using symlinks to mount it in `ui/media/modifier-thumbnails`. To make symlinks work I had to ensure fastapi is version 0.92.0 or greater.

image
image
image
image
image
image
image

### DIFF
--- a/scripts/on_env_start.bat
+++ b/scripts/on_env_start.bat
@@ -56,6 +56,10 @@ if "%update_branch%"=="" (
 @copy sd-ui-files\scripts\check_modules.py scripts\ /Y
 @copy "sd-ui-files\scripts\Start Stable Diffusion UI.cmd" . /Y
 @copy "sd-ui-files\scripts\Developer Console.cmd" . /Y
+if exist modifiers\ {
+    mklink ui\media\modifier-thumbnails\custom modifiers
+}
+
 
 @call scripts\on_sd_start.bat
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ installer
 installer.tar
 dist
 .idea/*
+ui/media/modifier-thumbnails/custom/
+ui/media/modifier-thumbnails/custom/*
+ui/media/modifier-thumbnails/custom/**/*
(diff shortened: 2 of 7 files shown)


scripts/on_env_start.bat
```diff
@@ -56,6 +56,10 @@ if "%update_branch%"=="" (
 @copy sd-ui-files\scripts\check_modules.py scripts\ /Y
 @copy "sd-ui-files\scripts\Start Stable Diffusion UI.cmd" . /Y
 @copy "sd-ui-files\scripts\Developer Console.cmd" . /Y
+if exist modifiers\ {
```
ogmaresca, line 59: These .bat scripts will need to be tested by someone on Windows. I only have Linux to test these changes

cmdr2:
Thanks! I suspect the .bat script needs some minor changes, will check this soon.

ogmaresca:
@cmdr2 I just updated the PR with a different way of getting custom modifier thumbnails without needing to upgrade fastapi. All of the shell scripts have been reverted

cmdr2:
Thanks @ogmaresca I'm reviewing this PR now

cmdr2:
@ogmaresca The code and design of the folder-structure looks great. Looks good to me. I'll test it out soon and will merge to beta if it goes well. Thanks! :)

cmdr2:
Merged, thanks!

PR #913, opened 2023-02-20
Fix the toggling of image modifiers

patriceac:
The toggling of image modifiers doesn't get properly applied if weights are changed after adding the image modifiers.

Before:
image

After:
image


patriceac:
Okay, not sure why my drawing got shifted. In the above screenshots I meant to show that 'Dot Art' is still present in the first screenshot (it shouldn't), and it's not included in the second one (desired behavior).

cmdr2:
Thanks!

PR #916, opened 2023-02-20
Don't scan safetensors files

JeLuF:
In newer versions of the picklescanner, scanning of .safetensors files creates an error:

21:28:01.067 ERROR MainThread ERROR: parsing pickle in D:\2.35\dev\models\stable-diffusion\dantionrealmix_10.safetensors: at position 1, opcode b'\xce' unknown

To avoid these entries in the logs, skip scanning of safetensors files.


cmdr2:
Thanks!

PR #917, opened 2023-02-20
Copy SSL DLLs

JeLuF:
Prevent the 'SSL module is not available' error message

https://discord.com/channels/1014774730907209781/1076822300214694008


cmdr2:
Wohoo! Thanks

PR #920, opened 2023-02-21
Fix the active selection's display

patriceac:
Yesterday's PR caused a regression on the active brush display, specifically for Sharpness, which is treated differently from the other brushes in the code. This is the fix.

Before:
image

After:
image


cmdr2:
Thanks!

PR #923, opened 2023-02-21
Download all: Add Metadata download

JeLuF:
image
- Popup to prevent accidental triggering of an image download
- Add ZIP download with JSON and folder support
- Use FileSaver polyfill for better browser support

https://discord.com/channels/1014774730907209781/1014774732018683928/1083307110361743441


ui/media/js/main.js
```diff
@@ -354,15 +356,27 @@ function onUseAsInputClick(req, img) {
...
     const imageInferenceSteps = img.getAttribute('data-steps')
     const imageGuidanceScale = img.getAttribute('data-guidance')
+    
+    return createFileName(imagePrompt, imageSeed, imageInferenceSteps, imageGuidanceScale, suffix)
+}
+
+function onDownloadJSONClick(req, img) {
+    const jsonDownload = document.createElement('a')
+    jsonDownload.download = getDownloadFilename(img, 'json')
+    jsonDownload.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(JSON.stringify(req, null, " ")))
```
cmdr2, line 371: Can you please check if this should be `req.reqBody`? I'm not sure, but would be good to confirm that we're encoding the correct JSON object.

JeLuF:
image

ui/media/js/main.js
JeLuF: The `req` object in the context of the button handler is the reqBody object, cf. https://github.com/cmdr2/stable-diffusion-ui/blob/beta/ui/media/js/main.js#L305:
```py
const req = Object.assign({}, reqBody, {
                seed: result?.seed || reqBody.seed
            })
```

cmdr2:
Thanks!

PR #924, opened 2023-02-22
Allow modifier images to be passed as base64 images

patriceac:
No change in existing UI behavior. This change allows image modifier plugins to (optionally) pass the modifier card image as a base64-encoded image.


cmdr2:
Thanks!

PR #925, opened 2023-02-22
Fix reloading of image tags with weight modifiers

patriceac:
Reloading of image tags with ((weight modifiers)) doesn't reuse the modifier card even if it exists, which means images are not restored either. This change fixes that behavior by ensuring proper matching of the tags with existing modifiers.

Original custom modifier:
image

Result when reloading weighted image tags:

Before:
image

After:
image


cmdr2:
Thanks!

PR #927, opened 2023-02-22
fix for thumbnail slider to support seamless resize

rbertus2000:
Just a little fix for the thumbnail slider to support live change of image size while changing the slider value.


cmdr2:
Thanks

PR #928, opened 2023-02-22
Fix up/down arrow keys on model selects

ogmaresca:
In https://github.com/cmdr2/stable-diffusion-ui/pull/894 I accidentally broke the up and down arrow keys switching between models in the dropdown. The issue was that the current logic requires the `<ul>` and `<li>` elements to be flat, but I changed it so the the element structure is a tree


cmdr2:
Thanks!

PR #930, opened 2023-02-23
Fix the display of the Preview options button

patriceac:
The preview options button overlaps the image task container when the window width is reduced because of the float:right property of the button. This technique makes the toolbar container grow as needed when it contains a floated element, resulting in cleaner display.

Before:
image

After:
image

To confirm, the UI displays as usual when the window is large enough (no change):
image

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -306,6 +306,7 @@
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #931, opened 2023-02-23
Fix the restoration of the inpainting toggle

patriceac:
The Inpainting toggle doesn't get restored at the very first attempt.

Repro steps:
1. Create a task with a source image and enable the inpainting toggle.
2. Copy the task to the clipboard
3. Refresh the page (F5)
4. Paste the task from the clipboard

Expected result:
The task gets restored and the toggle is ON.

Actual result:
The task is restored, but the toggle is OFF.

To fix that, we have to restore the toggle's state after loading the source image.

Before (upon initial task restoration):
image

After (upon initial task restoration):
image


cmdr2:
Thanks!

PR #932, opened 2023-02-24
DOM tweaking to identify modifier categories

patriceac:
This is purely a DOM update to be able to identify the category a given image modifier is part of - e.g. using .closest(). No UI change.

image

Sample code using this mechanism in an upcoming plugin:
const category = imageElement.closest('.custom-modifier-category').querySelector('h5').innerText.slice(1)


cmdr2:
Should this be called `modifier-category`? Isn't custom-modifier a different thing?

patriceac:
Good point. _Custom_ modifier categories are indeed a subset of modifier categories. Code updated.

cmdr2:
Thanks!

PR #935, opened 2023-02-25
Custom Image Modifiers dialog tweaks

patriceac:
Couple minor usability improvements for the custom image modifiers dialog:
- set the focus to the textbox when opening the dialog
- Escape key closes the dialog and cancels the changes, Ctrl+Enter confirms the changes. No change to the existing UI behavior using the mouse.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -387,7 +387,7 @@
         </div>
     </div>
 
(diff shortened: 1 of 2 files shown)


cmdr2:
Thanks!

PR #936, opened 2023-02-25
Refresh the image count as user types

patriceac:
Currently I have to change the focus for the image count to refresh. This change makes it immediate. I've been wondering if 'change' should merely be replaced by 'keyup' but decided against it for accessibility reasons (people who might be using accessibility tools with alternative input methods).

Hope this works for you.

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -1197,7 +1197,9 @@ function renameMakeImageButton() {
     }
 }
 numOutputsTotalField.addEventListener('change', renameMakeImageButton)
+numOutputsTotalField.addEventListener('keyup', renameMakeImageButton)
 numOutputsParallelField.addEventListener('change', renameMakeImageButton)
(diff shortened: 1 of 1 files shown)


JeLuF:
I'd like to suggest to debounce the key events so that the number only gets computed once the typing has stopped:

`numOutputsTotalField.addEventListener('keyup', debounce(renameMakeImageButton, 1000))`

patriceac:
The refreshing takes milliseconds or less, so I have to assume there's another motivation for only refreshing the count once the user stops typing. Can you help me understand where you're coming from? Not pushing back per se, just genuinely trying to understand the reasoning behind the suggestion.

JeLuF:
Yes, probably not really _required_ for this use case with only 2 or 3 digits being typed in. It's more of a best practice thing.

patriceac:
I made it 300ms to reduce the processing while limiting the lag. Thanks for the suggestion!

cmdr2:
Thanks!

PR #938, opened 2023-02-25
Make the image editor scrollable as needed

patriceac:
Some users on 4K screens zoom in their browser display, but this causes the editor buttons on the top right to be out of view, and since the editor is not scrollable, they have to zoom out to be able to hit 'save'.

This change fixes that by making the image editor window vertically scrollable if it gets too large. No UI change on smaller screens (e.g. phones).

Before:
image

After:
image


cmdr2:
Thanks!

PR #940, opened 2023-02-27
Select model by clicking on the file icon itself

patriceac:
Currently one has to click on the model name to select a model. Clicking on the file icon won't work and doesn't do anything. This change fixes that behavior by allowing the user to click on either the model name or the file icon to select a model.


cmdr2:
Thanks!

PR #945, opened 2023-02-28
Preserve full names for shortened modifiers

patriceac:
The PR https://github.com/cmdr2/stable-diffusion-ui/pull/779/files added code to preserve the full names of truncated image modifiers, but only in the "short image modifiers" code path. This PR fixes that by preserving the full image modifier in the DOM for both regular and truncated modifier names and making sure the right string is used (full string for comparisons, shortened string for display).


cmdr2:
Thanks!

PR #952, opened 2023-03-01
Image Modifiers Improvements

patriceac:
Same as previous. Proposing as a plugin for now for evaluation and feedback. We've already been discussing the whole concept of using hashtags to define custom categories. I appreciate you previously said you wanted something "more visual". I tried hard for several weeks to come up with something compelling, to no avail.

No matter how I tried to approach it with several prototypes, any attempt at a WYSIWYG UI brought me back to square one, and I found that text editing remains by far and large the easiest way to edit large lists of custom modifiers across several custom categories, so I thought I should bring it up again just in case.

This version supports custom modifier images, which is really a big plus IMHO, but I'll let you decide for yourself. The way to add the images is via drag & drop from any source, including generated images in the preview pane.

It also adds a search box to image modifiers as shown in the below screenshot.

Custom modifier images are added via drag & drop, either from the image generation pane of from external sources:
image

Search box for custom modifiers to quickly find what you're looking for:
image

Features:
    1. Allows for multiple custom modifier categories. Use # to name your custom categories, e.g.:
        #Custom category 1
        Custom modifier 1
        Custom modifier 2
        ...    
        #Custom category 2
        Custom modifier n
        ...
        #Custom category n...
        ...
    2. Restores the image modifier cards upon reloading the page.
    3. Adds a Clear All button on top of the task's image modifiers.
    4. Drop images on the modifier cards to set a custom visual (square pictures work best, e.g. 512x512).
    5. Adds import/export capabilities to the custom category editor (also exports/imports custom visuals).
    6. Collapses other categories when selecting a new one.
    7. Adds a search box for quick search of custom and predefined modifiers.

### DIFF
--- a/ui/plugins/ui/image-modifier-improvements.plugin.js
+++ b/ui/plugins/ui/image-modifier-improvements.plugin.js
@@ -192,6 +192,10 @@
                 }
                 // Call debounce function on filterImageModifierList function with 200ms wait time. Thanks JeLuf!
                 const debouncedFilterImageModifierList = debounce(filterImageModifierList, 200);
+
+                // add the searchbox
+                customModifierEntriesToolbar.insertAdjacentHTML('afterend', `<input type="text" id="image-modifier-filter" placeholder="Search for..." autocomplete="off"/>`)
+                imageModifierFilter = document.getElementById("image-modifier-filter") // search box
                 
                 // Add the debounced function to the keyup event listener
                 imageModifierFilter.addEventListener('keyup', debouncedFilterImageModifierList);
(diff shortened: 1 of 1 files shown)


patriceac:
> Hi there, I tested this in Firefox and noticed a few things:
> 
> * > The way to add the images is via drag & drop from any source, including generated images in the preview pane.
>   
>   
>   While it generally works from a file browser (nautilus) outside the browser window I can't add images from the preview pane - I assume that's the one on the right where generated images show up
> * Exporting the custom modifiers do not include the latest changes unless I close/unload and reopen the browser tab.

Thanks for reporting these issues. Both are fixed in this version, available as a plugin:

https://github.com/patriceac/Easy-Diffusion-Plugins/blob/main/plugins/Patrice/image-modifier-improvements.plugin.js

Bugfixes:
Fix image modifier d&d on Firefox
Export exactly what's shown in the textbox.

PR #956, opened 2023-03-03
Fix restoration of weighted tasks with truncated modifiers

patriceac:
Restoration of image tags comprising (((weighted modifiers with truncated nam...))) was broken because the code relied on the shorted modifier (with the ...) and added that to activeTags. On top of that, weights were lost during the restoration process because the code restored the non-weighted image modifiers.
Finally, modifiers passed as weighted enumerations (e.g. (({modifier1, modifier2, modifier3})) are parsed as ((modifier1)), (( modifier2), and (( modifier3)). Notice the extra spaces.

With this fix, reloading tasks and plugins restoring weighted and/or truncated modifiers should work as expected. Also opportunistically fixed the trimming of (({modifier1, modifier2, modifier3})) to properly compute as ((modifier1)), ((modifier2)), and ((modifier3)), i.e. without the extra spaces.


cmdr2:
Thanks!

PR #975, opened 2023-03-07
Add support for MPS when running on Apple silicon

michaelgallacher:
Changes:

* autodetect if MPS is available and if the pytorch version has MPS support.
* change logic from "is the device CPU?" to "is the device not CUDA?".
* set PYTORCH_ENABLE_MPS_FALLBACK=1

Known issues:

* Some samplers (eg DDIM) will fail on MPS unless forced to CPU-only mode


cmdr2:
Thanks!

PR #984, opened 2023-03-09
🔥 Hotfix 'torch.backend has no attribute MPS'

JeLuF:
https://discord.com/channels/1014774730907209781/1083420543405010994


michaelgallacher (approved):
I'm good with this but I don't understand the repro case.  This shouldn't happen if running with the version of torch we're using.

JeLuF:
> I'm good with this but I don't understand the repro case. This shouldn't happen if running with the version of torch we're using.

If you have an older isntallation of EasyDiffusion and kept upgrading it, it will use torch 1.11,

PR #987, opened 2023-03-09
Hotfix rollup

michaelgallacher:
* Reverts the recent 'torch.mps' changes since .mps is only available in torch v2.0, which isn't yet released.
* Includes Hotfitx 984
* Enables 'cpu-only' option when running on Apple silicon.

### DIFF
--- a/ui/easydiffusion/device_manager.py
+++ b/ui/easydiffusion/device_manager.py
@@ -35,7 +35,7 @@ def get_device_delta(render_devices, active_devices):
             render_devices = list(filter(lambda x: x.startswith("cuda:") or x == "mps", render_devices))
             if len(render_devices) == 0:
                 raise Exception(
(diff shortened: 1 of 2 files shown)


ui/easydiffusion/device_manager.py
```diff
@@ -204,7 +213,7 @@ def is_device_compatible(device):
         log.error(str(e))
         return False
 
-    if device == "cpu" or device == "mps":
+    if not is_cuda_available():
```
cmdr2, line 216: This isn't the same as `'cuda' in device`

cmdr2:
Thanks for the hotfix! I've merged the PR, and made the two changes to retain `device in ("cpu", "mps")` instead of only checking for CUDA availability. I do need to test CPU-compatibility occasionally, even though I have a CUDA-compatible card :)

PR #991, opened 2023-03-10
Fix autosave of modifiers image style

JeLuF:
https://discord.com/channels/1014774730907209781/1014774732018683926/1083821795427233973
image
Even though the image style gets saved, the setting doesn't get applied when loading the page.


cmdr2:
Thanks!

PR #992, opened 2023-03-10
Fix DDIM when running Apple MPS

michaelgallacher:
Copied a hack from a1111: force float32 tensor type on MPS.

This fixes the explicit and implicit (img2img) use of DDIM.


cmdr2:
Thanks! As discussed on Discord, I've applied this fix in sdkit, since it makes sense to fix it in the engine - https://github.com/easydiffusion/sdkit/commit/38e9edf591019b5ae5d6221fa8109cbb179c6071

(also added a check for `hasattr(torch.backends, "mps")` after that commit.

PR #1002, opened 2023-03-12
Plugin manager

patriceac:
An easy-to-use UI plugin manager. Pulls plugins from multiple repos, installs and updates plugins seamlessly, gives precedence to locally installed plugins (based on file names). Hope this one works for you.

If you want to test-drive it before merging, you can download the below as a plugin here: https://github.com/patriceac/Easy-Diffusion-Plugins/blob/main/plugin-manager.plugin.js

image


patriceac:
@cmdr2 just realized I inadvertently closed this PR while trying to cleanup my local GH repository. Is there any way to bring this back from the dead?

cmdr2:
@patriceac Ah I was wondering why this PR was closed :) The diffs are still visible in the PR, but the "reopen" button is disabled for me.

Might be easier to create a new PR? Otherwise I'll have to copy the changes by hand.

cmdr2:
I'm sure some git magic commands can bring it back from the dead. Nothing's ever lost, since the commit ids are still around.

But if you have the files around, that'll probably be a faster way to resurrect this.

patriceac:
Ping me when you're ready to proceed with the plugin manager integration and I'll create a new PR at that time. As I said above, the code is pretty self-contained so this should be quick and easy. And this will actually make my life easier in the meantime as this way I only have the plugin version to maintain for the time being.

cmdr2:
Sure, sounds good!

PR #1003, opened 2023-03-12
Notify plugins when image tags get refreshed

patriceac:
Image modifiers may be temporarily hidden by plugins like searchable modifier search box, and when restoring a task (e.g. use settings) the image modifier s copied with the hidden class set, which makes it look like it's missing.

By notifying plugins that the image tags have been refreshed, it allows it to act accordingly (in this case by making sure image tags are visible).

Before:
image

After:
image


cmdr2:
Thanks!

PR #1004, opened 2023-03-12
Warn users about small image dimensions

JeLuF:
Show a warning if the user chooses width & height < 512px

image


cmdr2:
Thanks!

PR #1005, opened 2023-03-12
Rename user hint

JeLuF:
Users place their plugins into the server plugin folder. To avoid this, rename the hint file.

https://discord.com/channels/1014774730907209781/1080996745762308226/1081001978412404757


cmdr2:
Thanks!

PR #1007, opened 2023-03-13
Allow creating custom HTML elements for image info button plugins

ogmaresca:
Right now the plugins for image info buttons can easily overflow, especially if you shrink image sizes. But there are some plugins that don't need to have so many lines, for example the OutpaintIt plugin needs 5 buttons when it could instead be 5 buttons on one line.

This makes two changes:
1. Add support for a `elem` function. This function has one parameter with the schema `{ createButton: () -> HTMLElement, req: Object, img: HTMLImgElement }`. `createButton` here is a function that takes a value from `IMAGE_INFO_BUTTONS`, while `req` and `img` are the same values from the `on_click` function.
2. The `text` field can now be a function that returns an `HTMLElement`, to support things like icons.

I didn't do anything about styling, so when using `elem` any text not created from the `createButton` function would need some custom background.

Example:

```js
(function() {
    function createElement(tagName, attributes, classes, text) {
        const element = document.createElement(tagName);
        if (attributes) Object.entries(attributes).forEach(([key, value]) => element.setAttribute(key, value));
        if (classes) classes.forEach(className => element.classList.add(className));
        if (text) element.appendChild(document.createTextNode(text));
        return element;
    }

    PLUGINS['IMAGE_INFO_BUTTONS'].push({
        text: () => createElement('i', undefined, ['fa-solid', 'fa-expand']),
    });

    PLUGINS['IMAGE_INFO_BUTTONS'].push({
        text: () => {
            const elem = createElement('span');
            elem.appendChild(document.createTextNode('This is '));
            elem.appendChild(createElement('strong', undefined, undefined, 'created with '));
            elem.appendChild(document.createTextNode(' a text function'));
            return elem;
        },
    });

    PLUGINS['IMAGE_INFO_BUTTONS'].push({
        elem: ({ createButton, req, img }) => {
            const elem = createElement('div', undefined, undefined);
            elem.appendChild(createElement(
                'span',
                { style: 'padding: 2px 4px; border-radius: var(--input-border-radius); background: var(--button-color); color: var(--button-text-color);' },
                undefined,
                'Example elem options: ',
            ))
            elem.appendChild(createButton({ text: 'Option 1' }));
            elem.appendChild(createButton({ text: 'Option 2' }));
            return elem;
        }
    });
})();
```

image

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -146,6 +146,9 @@ code {
 .imgItemInfo * {
     margin-bottom: 7px;
 }
+.imgItemInfo .tasksBtns {
+    margin-left: 1em;
+}
 .imgItem .image_clear_btn {
     transform: translate(40%, -50%);
 }
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -331,23 +331,27 @@ function showImages(reqBody, res, outputContainer, livePreview) {
             const imgItemInfo = imageItemElem.querySelector('.imgItemInfo')
             const img = imageItemElem.querySelector('img')
             const createButton = function(btnInfo) {
+                if (btnInfo.elem) {
+                    return btnInfo.elem({ createButton, req, img })
+                }
+
                 const newButton = document.createElement('button')
                 newButton.classList.add('tasksBtns')
-                newButton.innerText = btnInfo.text
+                newButton.appendChild(typeof btnInfo.text === 'function' ? btnInfo.text() : document.createTextNode(btnInfo.text))
                 newButton.addEventListener('click', function(event) {
                     btnInfo.on_click(req, img, event)
                 })
                 if (btnInfo.class !== undefined) {
                    newButton.classList.add(btnInfo.class)
                 }
-                imgItemInfo.appendChild(newButton)
+                return newButton
             }
             buttons.forEach(btn => {
                 if (btn.filter && btn.filter(req, img) === false) {
                     return
                 }
 
-                createButton(btn)
+                imgItemInfo.appendChild(createButton(btn))
             })
         }
     })
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -331,27 +331,59 @@ function showImages(reqBody, res, outputContainer, livePreview) {
             const imgItemInfo = imageItemElem.querySelector('.imgItemInfo')
             const img = imageItemElem.querySelector('img')
             const createButton = function(btnInfo) {
-                if (btnInfo.elem) {
-                    return btnInfo.elem({ createButton, req, img })
+                if (Array.isArray(btnInfo)) {
+                    const wrapper = document.createElement('div');
+                    btnInfo
+                        .map(createButton)
+                        .forEach(buttonElement => wrapper.appendChild(buttonElement))
+                    return wrapper
                 }
 
-                const newButton = document.createElement('button')
+                const isLabel = btnInfo.type === 'label'
+
+                const newButton = document.createElement(isLabel ? 'span' : 'button')
                 newButton.classList.add('tasksBtns')
-                newButton.appendChild(typeof btnInfo.text === 'function' ? btnInfo.text() : document.createTextNode(btnInfo.text))
-                newButton.addEventListener('click', function(event) {
-                    btnInfo.on_click(req, img, event)
-                })
+
+                if (btnInfo.html) {
+                    const html = typeof btnInfo.html === 'function' ? btnInfo.html() : btnInfo.html
+                    if (html instanceof HTMLElement) {
+                        newButton.appendChild(html)
+                    } else {
+                        newButton.innerHTML = html
+                    }
+                } else {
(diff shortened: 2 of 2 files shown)


cmdr2:
Thanks! I agree this is a growing problem that needs to be addressed.

Would this be a simpler, less-verbose developer experience?

Key aspects:
1. Allowing html in the `text` field of an object entry.
3. Allowing each "row" entry to be either an object entry (like it is today), or an array of objects. If it is an array of objects, then those elements will be stacked horizontally.
4. Allowing a new field called `type` in the object, which defaults to 'button'. If `type` is 'label`, then a `<label>` will be created instead.

As an alternative to aspect 1 - we could also allow a new `html` attribute, and preserve the current behavior of `text`. `text` would set `.innerText` and `html` would set `.innerHTML`. Which option would you prefer? Reusing the `text` field for text and html, or creating a new field called `html`, separately from `text`?

```
PLUGINS['IMAGE_INFO_BUTTONS'].push({ 'text': "Single button", on_click: foo }) // single button, like today

PLUGINS['IMAGE_INFO_BUTTONS'].push([
  { 'text': "<b>Button</b> 1", on_click: foo1 },
  { 'text': "Button 2", on_click: foo2 }
]) // two styled buttons in a row

PLUGINS['IMAGE_INFO_BUTTONS'].push([
  { 'text': "<b>Outpaint:</b>", type: 'label' },
  { 'text': "Button 1", on_click: foo1 },
  { 'text': "Button <em>2</em>", on_click: foo2 }
]) // a label and two styled buttons in a row
```

cmdr2:
I'm keen to use this for improving some of our core buttons as well, like giving more options to step counts etc. Thanks for your interest in this!

ogmaresca:
@cmdr2 I do like option 2, which can be implemented independently of any other change.

For an HTML string version of the entire row, I don't think that's a good idea. It would require global functions to be able to implement any event handlers, and making your own buttons would mean manual HTML instead of a function being provided to you.

For HTML button text, how about a `textType` field? I can also make it in general more intelligent with something like

```js
const textFactory = typeof btnInfo.text === 'function' ? btnInfo.text : () => btnInfo.text
const text = textFactory()
if (text instanceof HTMLElement) {
  newButton.appendChild(typeof btnInfo.text === 'function' ? btnInfo.text() : document.createTextNode(btnInfo.text))
} else if (btnInfo.textType === 'html') {
  newButton.innerHTML = text
} else {
  newButton.innerText = text
}
```

Then text can support these states:

1. `{ text: '<strong>Foo</strong>' }` would create a button with the text `&lt;strong&gt;Foo&lt;/strong&gt;`
2. `{ text: '<strong>Foo</strong>', textType: 'html' }` would create a button with the text **Foo**
3. `{ text: () => '<strong>Foo</strong>' }` would create a button with the text `&lt;strong&gt;Foo&lt;/strong&gt;`
4. `{ text: () => '<strong>Foo</strong>', textType: 'html' }` would create a button with the text **Foo**
5. `{ text: () => createElement({ tag: 'strong', text: 'Foo' }) }` would create a button with the text **Foo**

cmdr2:
@ogmaresca Sorry, I'm not sure I follow.

> For an HTML string version of the entire row, I don't think that's a good idea.

The suggestion in my previous message wasn't to put an HTML version of an entire row. Instead it suggested that a row could be represented via an array of objects. Please see the example below.

> It would require global functions to be able to implement any event handlers

I'm not sure where we'd need global functions or HTML strings. Working with an example (please see below), the snippet creates three rows. The first row contains a single button, the second row contains two buttons, the third row contains a text label and two buttons. Each button has an event handler attribute (`on_click`), which can be an anonymous function, or a locally-defined function.

Here's a modified suggestion, that allows the user to format the contents of a button/label via a separate `html` attribute (for things like bold etc). Otherwise they can use `text`. We continue to have an `on_click` handler (like we do today), to link to the handler function.

```js
PLUGINS['IMAGE_INFO_BUTTONS'].push({ 'text': "Single button", on_click: foo }) // single button, like today. this is what we do today

// this is a suggestion for two buttons in a single row. 
PLUGINS['IMAGE_INFO_BUTTONS'].push([
  { 'html': "<b>Button</b> 1", on_click: foo1 },
  { 'text': "Button 2", on_click: foo2 }
]) // two styled buttons in a row

// this is a suggestion for a text label, and then two buttons, all in a single row
PLUGINS['IMAGE_INFO_BUTTONS'].push([
  { 'html': "<b>Outpaint:</b>", type: 'label' },
  { 'text': "Button 1", on_click: foo1 },
  { 'html': "Button <em>2</em>", on_click: foo2 }
]) // a label and two styled buttons in a row
```

We could allow a separate field called `className` in the object, to allow people to style the element.

ogmaresca:
@cmdr2 I see what you're talking about. I updated the PR with changes to support your examples
image

cmdr2:
Thanks! I appreciate the change, and really like the implementation. Nice that `text` and `html` can also be functions :)

PR #1009, opened 2023-03-13
Only delete task sections if they are no longer processing

JeLuF:
When using the "close image"-X button, the task section gets removed when the last image in the section gets deleted. If the batch has not been completed yet, the remaining images get rendered to a hidden task section. 

This PR prevents the deletion of the task section while it is still processing.
https://discord.com/channels/1014774730907209781/1014774732018683928/1084938951271403550

### DIFF
--- a/ui/media/js/main.js
+++ b/ui/media/js/main.js
@@ -288,7 +288,10 @@ function showImages(reqBody, res, outputContainer, livePreview) {
                             allHidden = false;
                         }
                     }
-                    if(allHidden === true) {parentTaskContainer.classList.add("displayNone")}
+                    if(allHidden === true) {
+                        const req = htmlTaskMap.get(parentTaskContainer)
+                        if(!req.isProcessing || req.batchesDone == req.batchCount) {parentTaskContainer.classList.add("displayNone")}
+                    }
                 })
             })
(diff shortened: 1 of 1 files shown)


cmdr2:
I'm not sure about this one. I like having the ability to *immediately* remove bad tasks from the window, and don't mind a slight delay for the next task's processing. Bad tasks == tasks where the AI is going in the wrong direction.

How about we dispatch a "Stop task" event (to the server), along with hiding the section? That way, there'll be just a short delay (till the next sampling step, maybe 2-3 seconds) before it stops processing on the backend.

So this would achieve both goals - being able to remove tasks immediately, and preventing the server from completing the entire batch unnnecessarily.

JeLuF:
You can still delete the individual images with this PR.
Without this PR: You start a batch of 6 images. While it is rendering the first image, you delete the first image. The task gets hidden, while it is still rendering the 5 images you want to still look at. Reason for this is that the code only checked whether the task section is empty, without checking whether there are 5 images left to process.

Stopping the task is a good idea, but unrelated to the purpose of this PR.

cmdr2:
Okay, I agree. That makes sense.

PR #1013, opened 2023-03-14
Hardcode CSS content type

JeLuF:
By default, uvicorn uses the 'Content Type' value from 'Computer\HKEY_CLASSES_ROOT\.css' for the Content-Type header in its responses. Some systems have this set to 'application/x-css', an outdated content type used in the early days of CSS. Modern browsers ignore stylesheets that don't have the content-type 'text/css'. This change hardcodes the Content-Type to 'text/css', ignoring the registry.

https://discord.com/channels/1014774730907209781/1084874689769377892
https://discord.com/channels/1014774730907209781/1072804725017280553


cmdr2:
Thanks!

PR #1015, opened 2023-03-14
Fix #1012 - Image Modifiers > Thumbnail Size slider remembers its pos…

### LINKED ISSUE #1012: Image Modifiers > Thumbnail Size slider remembers its position but does not apply it at page load
**Describe the bug**
When using the Thumbnail Size slider in the Image Modifiers section, in subsequent sessions, the slider remembers the position it was set to. However the actual thumbnails are displayed at their default size regardless of the slider setting. You have to move the slider twice to make it apply a different setting, then go back to the one you want.

**To Reproduce**
1. Change the Thumbnail Size setting to a non-default value.
2. Reload the tab or open a new one.

**Expected behavior**
In addition to keeping the slider setting on reload, that setting should be applied to the thumbnails at page load, not only on value change.

 - OS: Linux Mint 21
 - Browser: Firefox 110.0.1
 - Version: 2.25.24

JeLuF:
…ition but does not apply it at page load

Loading of settings happens before the modifiers have been loaded, so the setting loader's triggering of 'change' events happens too early.

### DIFF
--- a/ui/media/js/image-modifiers.js
+++ b/ui/media/js/image-modifiers.js
@@ -149,10 +149,11 @@ async function loadModifiers() {
             createCollapsibles(editorModifierEntries)
         }
     } catch (e) {
-        console.log('error fetching modifiers', e)
+        console.error('error fetching modifiers', e)
     }
 
     loadCustomModifiers()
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #1018, opened 2023-03-15
remove task from dome when removing images

AssassinJN:
Changing the functionality to remove task from dom when all images have been removed. This will save system memory in the browser allowing better performance.


cmdr2:
@JeLuF - any reason not to do this? We can always change this logic later, when implementing undo? Thanks

JeLuF:
This section will have to be rewritten for undo anyway. Shouldn't harm to change this.

cmdr2:
Thanks!

PR #1019, opened 2023-03-15
When zooming, only show one copy of the image

JeLuF:
Don't add a copy of the image for each live preview image shown:
image


cmdr2:
Thanks!

PR #1020, opened 2023-03-16
Support custom folder filename formats

ogmaresca:
Add two new settings: folder format and filename format. They both support placeholders for the prompt, seed, and datetime components. The filenames also support the image number in the folder and two representations of the timestamp.

The default values should generate the exact same values that are currently being generated. When loading the settings, if the format is the default value the default radio button will be checked. When the custom radio button is checked, it'll show an input field and a list of supported placeholders, and when the default is checked both will be hidden. The folder format does support seeds, minutes, and seconds, but I didn't include that in the UI for folders.

image
image
image

### DIFF
--- a/ui/easydiffusion/types.py
+++ b/ui/easydiffusion/types.py
@@ -45,6 +45,9 @@ class TaskData(BaseModel):
     stream_image_progress: bool = False
     stream_image_progress_interval: int = 5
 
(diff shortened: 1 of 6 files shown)


cmdr2:
Thanks @ogmaresca! I will get to this PR soon.

PR #1034, opened 2023-03-19
Make file names of downloaded files shorter

JeLuF:
Long file and folder names cause problems on Windows systems.


cmdr2:
Thanks!

PR #1036, opened 2023-03-19
Prevent collisions of 'What's new' and the server status

JeLuF:
With the current setting of 1265px, the status and the tab headers overlap:
image


cmdr2:
Thanks!

PR #1038, opened 2023-03-20
Add 'Karras' tag to DPM samplers

JeLuF:
image

Alternatively, `<optgroup>` could be used, but it uses more space, adding a scrollbar to the dropdown:
image


cmdr2:
Let's go with this approach, thanks! :)

PR #1041, opened 2023-03-20
Add undo support

JeLuF:
Make image deletion, task deletion and automatic task deletion due to deletion of its last image undoable.
The easiest way to implement the undo of the "automatic task deletion due to deletion of its last image" was to use a stack of undoable actions. To limit memory consumption, the undo buffer is limited to 20 entries. 

When elements get evicted from the undo buffer, the related entries in `htmlTaskMap` and `imageRequests` are deleted to reduce memory consumption.

image
The preview toolbar is kept visible if there are undoable actions.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -282,18 +282,16 @@
         </div>
 
         <div id="preview" class="col-free">
-            <div id="initial-text">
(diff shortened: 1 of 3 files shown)


cmdr2:
Thanks! Much appreciated ;)

cmdr2:
I added support for Ctrl+Z as well

PR #1045, opened 2023-03-21
Add LORA model to metadata files

ogmaresca:
`lora_alpha` is commented out in `TASK_TEXT_MAPPING` because otherwise it would save as `LoRA Strength: 0` instead of `LoRA Strength: 100`. When lora strength is implemented (hopefully soon!) it should be uncommented.

If `test_diffusers` is disabled, then `LoRA model` will be removed instead of showing `LoRA model: None`


cmdr2:
Thanks! We may need a corresponding change in `ui/js/dnd.js` to allow this metadata change to be loaded into the UI.

PR #1048, opened 2023-03-22
🔥Hotfix: Avoid name clashes for autosave files

JeLuF:
The old code had a 1.6% chance of name collisions (AAb/345 and AAB3/45 would generate the same filename) on Linux and 39% on Windows (base64 is case sensitive, Windows isn't). This has already led to data loss: https://discord.com/channels/1014774730907209781/1087720718034280519

This code uses base36 (0-9, A-Z) to avoid case sensitivity issues. To avoid collisions on fast computers or multi GPU computers, the time resolution of the timestamp is changed from seconds to 0.1ms, and the image number isn't added to the timestamp but appended as an extra character. Due to the limitation of the timestamps to 7 characters, the timecode will repeat every 90 days. This shouldn't be an issue since most sessions will not last this long.

Due to the potential data loss and the probably long time that it will take before the next release, this is a PR for the main branch, not to the beta branch.

### DIFF
--- a/ui/easydiffusion/utils/save_utils.py
+++ b/ui/easydiffusion/utils/save_utils.py
@@ -1,11 +1,11 @@
 import os
 import time
-import base64
 import re
 
 from easydiffusion.types import TaskData, GenerateImageRequest
 
 from sdkit.utils import save_images, save_dicts
+from numpy import base_repr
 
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks! Appreciate this investigation and fix!

PR #1051, opened 2023-03-24
Fix the willReadFrequently warning on Chrome

patriceac:
Fix this warning on Chrome when saving a drawing in the image editor:
image

More info: https://stackoverflow.com/questions/74101155/chrome-warning-willreadfrequently-attribute-set-to-true

### DIFF
--- a/ui/media/js/image-editor.js
+++ b/ui/media/js/image-editor.js
@@ -414,7 +414,7 @@ class ImageEditor {
 			this.layers[name] = {
 				name: name,
 				canvas: canvas,
-				ctx: canvas.getContext("2d")
+				ctx: canvas.getContext("2d", { willReadFrequently: true })
 			}
 		})
 


cmdr2:
Thanks, do we read from the canvas frequently?

patriceac:
Does it justify the additional vram memory consumption? Not sure. I've been hesitant to submit this one; I'll trust your judgement. Merge it or close it as you see fit.

cmdr2:
Thanks! I'd rather keep things simple, because a single image doesn't consume much VRAM. And we don't read/write frequently.

The VRAM requirements for a regular image is very different from a Stable Diffusion image, because an SD image works in a different way. A single regular image doesn't consume much VRAM, think about the number of textures present in a simple game. A single game scene will often have tens to hundreds textures, each varying from 512x512 upto 4096x4096.

PR #1052, opened 2023-03-24
Fix for Clear button in image editor

patriceac:
Repro steps:
From a fresh page, click 'Draw'. Image editor launches.
Draw something and hit 'Save'.
Click 'Draw' again, then click 'Clear'

Expected behavior:
The canvas is cleared.

Actual result:
The previous drawing remains there.

This is happening because during the Save process, the drawing is copied to the background canvas, but the latter is not properly cleared by the Clear button.


cmdr2:
Thanks!

PR #1062, opened 2023-03-25
Fix restoration of disabled image tags

patriceac:
Fixes a regression introduced by PR 1003 that causes _disabled_ image tags to be restored in an _enabled_ state by the "image modifiers improvements" plugin. No change to the regular UI.


cmdr2:
Thanks!

PR #1075, opened 2023-03-27
Reworked the task entry config section

lucasmarcelli:
https://github.com/cmdr2/stable-diffusion-ui/pull/1074 closed this PR so i could go against the beta branch


lucasmarcelli:
Added: 
- Setting for deciding which config should show
- Proper save support for multiselect settings
- Reusable way to add more task configs easily, adding to `getTaskConfigSetup` in `ui\media\js\parameters.js` will add it both to settings and the task card

Screenshot 2023-03-27 234232

Screenshot 2023-03-27 234458

cmdr2:
Hi @lucasmarcelli Thanks for the PR! I think the only field missing today (in the task header) is `Dimensions`. The rest of the missing fields are not image-specific, for e.g. the image doesn't change based on the `VRAM Usage Level`, or `Output format` etc. So they don't need to be displayed in the task header section.

The task header is meant to let a user know (quickly) what the images in the task will look like. For e.g. if they have a lot of tasks, and they've minimized the tasks (by clicking on the header). They can glance through all the minimized tasks to get an idea about each task.

So specifically, I'd include the `Dimensions` field in the task header, since that's often useful to know quickly. Otherwise, sorry but I don't think the added complexity is justified. You're always free to write a plugin for yourself, if it's something you wish to customize.

I definitely appreciate the effort, and sorry we couldn't accept this PR. Thanks! :)

cmdr2:
Looking through the code one more time, I like the refactoring of the task info into an object (with generator functions for each setting).

So if you're interested, I'd definitely like to merge that bit of refactoring. The current code (in our codebase) for generating the config by appending HTML strings is pretty hacky :)

ui/media/js/main.js
```diff
@@ -960,33 +984,8 @@ function createTask(task) {
...
-        taskConfig += `, <b>Hypernetwork:</b> ${task.reqBody.use_hypernetwork_model}`
-        taskConfig += `, <b>Hypernetwork Strength:</b> ${task.reqBody.hypernetwork_strength}`
-    }
-    if (task.reqBody.use_lora_model) {
-        taskConfig += `, <b>LoRA:</b> ${task.reqBody.use_lora_model}`
-    }
-    if (task.reqBody.preserve_init_image_color_profile) {
-        taskConfig += `, <b>Preserve Color Profile:</b> true`
-    }
+    taskConfig += `<div class="taskConfigData">${createTaskConfig(task)}</span></div></div>`;
```
cmdr2, line 988: Nice!

ui/media/js/parameters.js
```diff
@@ -198,8 +198,71 @@ var PARAMETERS = [
...
+            'negative_prompt',
+            'prompt_strength',
+            'use_face_correction',
+            'upscale',
+            'use_hypernetwork_model',
+            'hypernetwork_strength',
+            'use_lora_model',
+            'preserve_init_image_color_profile'
+        ],
+        taskConfig: {
```
cmdr2, line 231: I really like this pattern!

lucasmarcelli:
will open a new PR

PR #1078, opened 2023-03-28
Automatic AMD GPU detection on Linux

DianaNites:
Automatically detects AMD GPUs and installs the ROCm version of PyTorch instead of the cuda one

A later improvement may be to detect the GPU ROCm version and handle GPUs that dont work on upstream ROCm, ether because they're too old and need a special patched version, or too new and need `HSA_OVERRIDE_GFX_VERSION=10.3.0` added, possibly check through `rocminfo`?

For AMD this uses the current latest ROCm version of 5.4.2, and PyTorch 2.0 which is only 2.0 because they felt like it and is basically 1.14 and is entirely backwards compatible, per [their FAQ](https://pytorch.org/get-started/pytorch-2.0/#faqs)

nvidia/cuda is unchanged

----

This was tested and working on a up to date as of this comment Arch Linux system, with ROCm installed according to <https://wiki.archlinux.org/title/GPGPU#ROCm>, with a AMD RX 6800M, GFX1031, with `HSA_OVERRIDE_GFX_VERSION=10.3.0` hard-coded locally.

### DIFF
--- a/scripts/on_sd_start.sh
+++ b/scripts/on_sd_start.sh
@@ -73,11 +73,11 @@ DGPU_BRAND=$(lspci 2>/dev/null | grep Display)
 if python ../scripts/check_modules.py torch torchvision; then
     # temp fix for installations that installed torch 2.0 by mistake
     if [ "$OS_NAME" == "linux" ]; then
-        # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
-        # and if they do they can edit this themselves like AMD users already have to
-        if echo "$DGPU_BRAND" | grep -q "AMD"; then
-            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
-        elif echo "$GPU_BRAND" | grep -q "AMD"; then
+        # Check for AMD and NVIDIA dGPUs, always preferring an NVIDIA GPU if available
+        # Fall back to NVIDA/CUDA if somehow none of these checks pass
+        if echo "$DGPU_BRAND" | grep -q "NVIDIA" || echo "$GPU_BRAND" | grep -q "NVIDIA"; then
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        elif echo "$DGPU_BRAND" | grep -q "AMD" || echo "$GPU_BRAND" | grep -q "AMD"; then
             python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
         else
             python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
@@ -94,11 +94,11 @@ else
     export PYTHONPATH="$INSTALL_ENV_DIR/lib/python3.8/site-packages"
 
     if [ "$OS_NAME" == "linux" ]; then
-        # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
-        # and if they do they can edit this themselves like AMD users already have to
-        if echo "$DGPU_BRAND" | grep -q "AMD"; then
-            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" || fail "Installation of torch and torchvision for ROCm failed"
-        elif echo "$GPU_BRAND" | grep -q "AMD"; then
+        # Check for AMD and NVIDIA dGPUs, always preferring an NVIDIA GPU if available
+        # Fall back to NVIDA/CUDA if somehow none of these checks pass
+        if echo "$DGPU_BRAND" | grep -q "NVIDIA" || echo "$GPU_BRAND" | grep -q "NVIDIA"; then
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 || fail "Installation of torch and torchvision for CUDA failed"
+        elif echo "$DGPU_BRAND" | grep -q "AMD" || echo "$GPU_BRAND" | grep -q "AMD"; then
             python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" || fail "Installation of torch and torchvision for ROCm failed"
         else
             python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 || fail "Installation of torch and torchvision for CUDA failed"
--- a/scripts/on_sd_start.sh
+++ b/scripts/on_sd_start.sh
@@ -97,18 +97,13 @@ else
         # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
         # and if they do they can edit this themselves like AMD users already have to
         if echo "$DGPU_BRAND" | grep -q "AMD"; then
-            installed=$(python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2")
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" || fail "Installation of torch and torchvision for ROCm failed"
         elif echo "$GPU_BRAND" | grep -q "AMD"; then
-            installed=$(python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2")
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" || fail "Installation of torch and torchvision for ROCm failed"
         else
-            installed=$(python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116)
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 || fail "Installation of torch and torchvision for CUDA failed"
         fi
-        
-        if [ -n "$installed" ]; then
-            echo "Installed."
-        else
-            fail "torch install failed"
-        fi
+        echo "Installed."
     elif [ "$OS_NAME" == "macos" ]; then
         if python -m pip install --upgrade torch==1.13.1 torchvision==0.14.1 ; then
             echo "Installed."
--- a/scripts/on_sd_start.sh
+++ b/scripts/on_sd_start.sh
@@ -63,11 +63,25 @@ case "${OS_NAME}" in
     *)          echo "Unknown OS: $OS_NAME! This script runs only on Linux or Mac" && exit
 esac
 
+# This is the "main" GPU, but some systems will have extra GPUs as "Display Controllers"
+# Notably, AMD laptops with a AMD dGPU, but also probably any desktop with a second GPU.
+# For nvidia it already works fine
+GPU_BRAND=$(lspci 2>/dev/null | grep VGA)
+DGPU_BRAND=$(lspci 2>/dev/null | grep Display)
+
 # install torch and torchvision
 if python ../scripts/check_modules.py torch torchvision; then
     # temp fix for installations that installed torch 2.0 by mistake
     if [ "$OS_NAME" == "linux" ]; then
-        python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
+        # and if they do they can edit this themselves like AMD users already have to
+        if echo "$DGPU_BRAND" | grep -q "AMD"; then
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
+        elif echo "$GPU_BRAND" | grep -q "AMD"; then
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
+        else
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        fi
     elif [ "$OS_NAME" == "macos" ]; then
         python -m pip install --upgrade torch==1.13.1 torchvision==0.14.1 -q
     fi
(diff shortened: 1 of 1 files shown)


scripts/on_sd_start.sh
```diff
@@ -63,11 +63,25 @@ case "${OS_NAME}" in
     *)          echo "Unknown OS: $OS_NAME! This script runs only on Linux or Mac" && exit
 esac
 
+# This is the "main" GPU, but some systems will have extra GPUs as "Display Controllers"
+# Notably, AMD laptops with a AMD dGPU, but also probably any desktop with a second GPU.
+# For nvidia it already works fine
+GPU_BRAND=$(lspci 2>/dev/null | grep VGA)
```
JeLuF, line 69: lspci is not included in the standard installation of many Linux distributions.

scripts/on_sd_start.sh
```diff
@@ -63,11 +63,25 @@ case "${OS_NAME}" in
...
-        python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
+        # and if they do they can edit this themselves like AMD users already have to
+        if echo "$DGPU_BRAND" | grep -q "AMD"; then
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
+        elif echo "$GPU_BRAND" | grep -q "AMD"; then
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" -q
+        else
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        fi
```
JeLuF, line 84: Multi-GPU setups will get the (slower) ROCm installed, even if a more powerfull NVidia card would also be present.

scripts/on_sd_start.sh
```diff
@@ -80,7 +94,17 @@ else
     export PYTHONPATH="$INSTALL_ENV_DIR/lib/python3.8/site-packages"
 
     if [ "$OS_NAME" == "linux" ]; then
-        if python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 ; then
+        # Check for AMD dGPU first, assume nobody has a primary nvidia GPU and AMD secondary GPU,
+        # and if they do they can edit this themselves like AMD users already have to
+        if echo "$DGPU_BRAND" | grep -q "AMD"; then
+            installed=$(python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2")
```
JeLuF, line 100: Suppressing the output of pip is making debugging of user errors more difficult. Collecting the return code instead of stdout should be sufficient.

scripts/on_sd_start.sh
```diff
@@ -80,7 +94,17 @@ else
...
+        elif echo "$GPU_BRAND" | grep -q "AMD"; then
+            installed=$(python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2")
+        else
+            installed=$(python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116)
+        fi
+        
+        if [ -n "$installed" ]; then
```
JeLuF, line 107: In case of a download failure, there is stdout output but the installation did fail. Consider:
```sh
python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2" || fail "Installation of torch and torchvision for rocm failed"
```

scripts/on_sd_start.sh
DianaNites: This could be done without it then, by checking `/sys/class/drm/card?/../../vendor` for the AMD or NVIDIA vendor ID, but that would be more complicated bash scripting i'm not equipped  to do.

scripts/on_sd_start.sh
DianaNites: Would detecting if theres an AMD *or* NVIDIA dGPU, and always prioritizing an NVIDIA one, be acceptable?

scripts/on_sd_start.sh
DianaNites: technically, as i understand the Linux sysfs rules, the most correct way to do this manually with sysfs is to walk up the path from the `card?` path, checking the subsystem of each parent directory, and halting when its the expected subsystem type(in this case, `pci`), because even though theres a `device` link, documentation explictly calls out using it as an application bug.

DianaNites: oh, yeah. https://www.kernel.org/doc/html/latest/admin-guide/sysfs-rules.html

> You need to walk up the chain until you find the device that matches the expected subsystem. Depending on a specific position of a parent device or exposing relative paths using ../ to access the chain of parents is a bug in the application.

to be correct it needs to walk

scripts/on_sd_start.sh
cmdr2: Yes, I think prioritizing an NVIDIA GPU (over AMD) is a good compromise for now. We can revisit this in the future if it becomes an issue.

DianaNites:
Besides the lspci issue, current review comments have been addressed.
I'm not confident enough in my bash shell scripting to properly implement the sysfs walk though

all it needs to do is find every `card` in `/sys/class/drm` and walk up the resolved path, checking for the existence of a `subsystem` link, read its target, and check if its last element is "pci" to find the actual GPU device, where it can then read the `vendor` file and if its the string `0x10de`, its NVIDIA, and if its `0x1002` its AMD (see `/usr/share/hwdata/pci.ids` / `pci.ids(5)` for reference)

JeLuF:
I've thought about the lspci/sysfs detection a bit more yesterday evening. The code looks for the card, but it doesn't check for a ROCm installation. Is there an easy way to see whether the Linux ROCm driver is working?

DianaNites:
Not that I know of. How is it handled for checking for cuda?

JeLuF:
Up to now, we just assumed CUDA would be available, or fall back to CPU, which is provided by the CUDA versions of torch/torchvision.

```bash
if which nvidia-smi; then
    nvidia-smi -L;
fi

if which rocminfo; then
    rocminfo |grep Marketing.Name
fi
```

returns:
```
/usr/bin/nvidia-smi
GPU 0: NVIDIA GeForce RTX 3060 (UUID: GPU-c36e688b-893e-1342-fc7b-a9dff0a0680d)
GPU 1: NVIDIA GeForce RTX 3060 (UUID: GPU-c654638f-7c6c-ee43-b096-be1af1274f90)
/usr/bin/rocminfo
  Marketing Name:          12th Gen Intel(R) Core(TM) i5-12500
  Marketing Name:          AMD Radeon RX 6600
```

`rocminfo` could be a good way to detect an actual ROCm installation.

DianaNites:
I dont think checking `rocmino` will be reliable, Arch Linux at least seems to intentionally not put `/opt/rocm/bin` in `$PATH`, and I think thats the right move because when i was trying to get AMD to work, some ROCm versions of opencl tools for example *will crash your system*, `/opt/rocm/clinfo` someehow maaged to do this for me a few times. I don't know what other distros do. Automatically handling this cross-distro would probably be non-trivial.

I think it's worthwhile to get AMD up to par with NVIDIA, with this PR first before making the installer even smarter. Does automatically detecting/installing ROCm/CUDA have to block this PR?

JeLuF:
Using `lspci` is a blocker for me.
```
jf@Monster:~/work$ grep amdgpu /proc/bus/pci/devices
0600    100273ff        99            420000000c                       0              410000000c                       0                    4001                53b00000                53c00000               200000000                0                  200000                       0                     100                  100000                   20000        amdgpu
jf@Monster:~/work$ grep nvidia /proc/bus/pci/devices
0100    10de2487        9b              52000000              602000000c                       0              603000000c                       0                    6001                53000000                 1000000         10000000                       0                 2000000                       0                      80                   80000        nvidia
0800    10de2487        9c              50000000              600000000c                       0              601000000c                       0                    3001                51000000                 1000000         10000000                       0                 2000000                       0                      80                   80000        nvidia
```
Could we use this to detect the installed GPUs? I made a PR to your branch to show what I mean.
https://github.com/DianaNites/stable-diffusion-ui/pull/1

JeLuF:
I've tried EasyDiffusion with ROCm on Ubuntu and an RX 6600. I upgraded torch using
```
python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2"
```
and encounter two errors in the logs. This one is repeating continuously:
```
10:07:15.321 ERROR MainThread HIP error: invalid device ordinal   - device_manager.py:230
HIP kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.                                                                   
For debugging consider passing HIP_LAUNCH_BLOCKING=1.                                                                                                                                    
Compile with `TORCH_USE_HIP_DSA` to enable device-side assertions.                                                                                                                       
```
And this is what I got when rendering an image:
```
10:09:27.498 ERROR cuda:0 Traceback (most recent call last):                                                                                                                                  task_manager.py:346
  File "/home/jf/work/easy-diffusion-ROCm/ui/easydiffusion/task_manager.py", line 336, in thread_render
    task.response = renderer.make_images(
  File "/home/jf/work/easy-diffusion-ROCm/ui/easydiffusion/renderer.py", line 38, in make_images
    images, seeds = make_images_internal(req, task_data, data_queue, task_temp_images, step_callback)
  File "/home/jf/work/easy-diffusion-ROCm/ui/easydiffusion/renderer.py", line 59, in make_images_internal
    images, user_stopped = generate_images_internal(
  File "/home/jf/work/easy-diffusion-ROCm/ui/easydiffusion/renderer.py", line 91, in generate_images_internal
    images = generate_images(context, callback=callback, **req.dict())
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/sdkit/generate/image_generator.py", line 46, in generate_images
    seed_everything(seed)
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/pytorch_lightning/utilities/seed.py", line 69, in seed_everything
    torch.manual_seed(seed)
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/torch/random.py", line 40, in manual_seed
    torch.cuda.manual_seed_all(seed)
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/torch/cuda/random.py", line 113, in manual_seed_all
    _lazy_call(cb, seed_all=True)
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/torch/cuda/__init__.py", line 183, in _lazy_call
    callable()
  File "/home/jf/work/easy-diffusion-ROCm/installer_files/env/lib/python3.8/site-packages/torch/cuda/random.py", line 110, in cb
    default_generator = torch.cuda.default_generators
IndexError: tuple index out of range
```
Full log: [amd.txt](https://github.com/cmdr2/stable-diffusion-ui/files/11109414/amd.txt)

@DianaNites , do you have any tips?

DianaNites:
AMD 6600 or 6600 XT? It might make a difference

Unfortunately, AMD ROCm sometimes needs some hand-holding to work because only up to gfx1030 is "officially" supported out of the box.

The 6600XT is gfx1032, but should work, like my gfx1031 card, with `HSA_OVERRIDE_GFX_VERSION=10.3.0` set. This is a ROCm issue https://github.com/RadeonOpenCompute/ROCm/issues/1698#issuecomment-1261646854.

I dont know if theres any good way to automatically detect needing this workaround on easy diffusion's side, always adding it wont work because then it will break AMD cards that are *older* than gfx1030, some of which *are* officially supported and would work out of the box. Hopefully AMD fixes this someday...

It's non-optimal, but at least its a really easy to explain troubleshooting step, "if AMD doesnt work out of the box, try `HSA_OVERRIDE_GFX_VERSION=10.3.0 stable-diffusion-ui-server` instead", and after that all the easy diffusion easy goodness should Just Work

DianaNites:
I've merged your `/proc/bus/devices` patch, that should work nicely.

DianaNites:
@cmdr2 I can change it to merge to the beta if you want? Besides that If it looks good for you, then I dont think theres anything left for initial support.

scripts/on_sd_start.sh
```diff
@@ -63,11 +63,30 @@ case "${OS_NAME}" in
...
 # install torch and torchvision
 if python ../scripts/check_modules.py torch torchvision; then
     # temp fix for installations that installed torch 2.0 by mistake
     if [ "$OS_NAME" == "linux" ]; then
-        python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116 -q
+        # Check for AMD and NVIDIA dGPUs, always preferring an NVIDIA GPU if available
+        if [ "$HAS_NVIDIA" != "yes" -a "$HAS_AMD" = "yes" ]; then
+            python -m pip install --upgrade torch torchvision --extra-index-url "https://download.pytorch.org/whl/rocm5.4.2"  || fail "Installation of torch and torchvision for AMD failed"
+        else
+            python -m pip install --upgrade torch==1.13.1+cu116 torchvision==0.14.1+cu116 --extra-index-url "https://download.pytorch.org/whl/cu116"  || fail "Installation of torch and torchvision for CUDA failed"
```
cmdr2, line 88: Can you please add back the `-q` quiet flag? These lines are run each time the program starts up, to force the package back to the specified version of torch.

cmdr2:
Yes thanks! Please change it to `beta`, and I'll merge it.

The PR looks good to me (just left a comment about the `-q` flag).

DianaNites:
`-q` flag has been added back, good catch, and and the PR should now go into `beta`

DianaNites:
Hmm, one thing re: `-q` flag

If a user with an AMD GPU has an existing install, and were using the CPU before, then when they switch, it will be through the *already installed* path, meaning it will appear to freeze as pip *quietly* installs the ROCm version of torch

Is there a good solution to this?

cmdr2:
@DianaNites  Hmm good point. A shell script solution would be to grab the current version from `pip show torch` (via grep or if pip can be convinced to print just the version), compare it against the desired version. If different, print that we're installing something. We can then get rid of the `-q` flag.

This PR is only for Linux, so we don't need to touch the .bat scripts.

scripts/on_sd_start.sh
JeLuF: Please don't run stuff quietly. Makes it really hard if you have to debug stuff that you can't see.

cmdr2:
@DianaNites  I have a small tweak coming up in beta (for the torch installation step), which should avoid the "silent install and freeze" problem entirely. Please hold on. Thanks

DianaNites:
Okay, can do.

andreas-it-dev-llc:
pls merge

cmdr2:
Thanks @DianaNites and @JeLuF ! Merged, and will integrate this into the updated script (that I'm working on).

cmdr2:
@DianaNites I've integrated your change into the new package management script. Does this look okay to you? And does it work on your PC?

https://github.com/cmdr2/stable-diffusion-ui/blob/beta/scripts/check_modules.py#L82

The thing missing from the new script is automatically upgrading non-ROCm users to the ROCm version. This needs to be done, but I won't be able to get to it immediately.

New installations will automatically get the ROCm version (if running AMD on Linux).

Thanks again for your help here! :)

cmdr2:
Thanks @DianaNites ! Yeah, it should download sd-v1-4.ckpt in a fresh install, will check. The change is to allow users to delete sd-v1-4, as long as there's atleast one model in the folder tree. Sounds like a bug, thanks for reporting!

I'm glad the new change is working for you!

I was wondering if you have any ideas for this error - `RuntimeError: Input type (c10::Half) and bias type (float) should be the same`. This was reported by a user (AMD/Linux) who started using the ROCm version yesterday.

I've attached the log file. I'm not sure if you're on Discord, but here's the relevant thread: https://discord.com/channels/1014774730907209781/1014780368890630164/1097848084551172117

**Steps to reproduce (for that user):** fresh install on AMD/Linux, switch to beta, uninstall torch/torchvision via dev-console, run `./start.sh` to get the ROCm version, then generate an image with any sampler (PLMS) and sd-v1-4.

It's surprising because logs show that the first "warm-up" image generation works okay (it happens right after a model loads, by running a 1-step PLMS for a 64x64 image). Subsequent user-initiate image generations seem to fail.

Any ideas welcome! Thanks again for all your help!

Log file: [message (55).txt](https://github.com/cmdr2/stable-diffusion-ui/files/11269716/message.55.txt)

PR #1087, opened 2023-03-30
Allow loading/saving app.config from plugins and support custom folder/filename formats from app.config

ogmaresca:
Sequel to https://github.com/cmdr2/stable-diffusion-ui/pull/1020

* Add a listener for `PARAMETERS.push` to add new setting rows when parameters are added
* Support setting parameter values and saving to app.config if `saveInAppConfig`. The `extra=Extra.allow` in `SetAppConfigRequest` is what allows the Python code to retrieve the extra properties
* Support `filename_format` and `folder_format` configs from app.config. The default values will generate the exact same format that it currently does

I added [file-folder-format.plugin.js](https://github.com/ogmaresca/easydiffusion-plugins/blob/main/ui/file-folder-format.plugin.js) to use the new functionality

### DIFF
--- a/ui/easydiffusion/utils/save_utils.py
+++ b/ui/easydiffusion/utils/save_utils.py
@@ -1,13 +1,18 @@
 import os
 import time
+import base64
 import re
 
+from easydiffusion import app
 from easydiffusion.types import TaskData, GenerateImageRequest
+from functools import reduce
+from datetime import datetime
 
 from sdkit.utils import save_images, save_dicts
 from numpy import base_repr
 
 filename_regex = re.compile("[^a-zA-Z0-9._-]")
+img_number_regex = re.compile("([0-9]{5,})")
 
 # keep in sync with `ui/media/js/dnd.js`
 TASK_TEXT_MAPPING = {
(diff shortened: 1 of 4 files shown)


ui/easydiffusion/utils/save_utils.py
```diff
@@ -31,12 +36,86 @@
...
+            self._evaluated = self._factory()
+        return self._evaluated
+
+def format_placeholders(format: str, req: GenerateImageRequest, task_data: TaskData, now = None):
+    if now is None:
+        now = time.time()
+
+    for placeholder, time_format in time_placeholders.items():
+        if placeholder in format:
+            format = format.replace(placeholder, datetime.fromtimestamp(now).strftime(time_format))
```
cmdr2, line 72: Thanks for doing it this way. Safer than executing arbitrary user-specified format strings.

ui/media/js/parameters.js
```diff
@@ -243,29 +254,74 @@ function getParameterElement(parameter) {
...
+        const note = typeof parameter.note === 'function' ? parameter.note(parameter) : parameter.note
+        const noteElements = []
+        if (note) {
+            const noteElement = createElement('small')
+            if (note instanceof Node) {
+                noteElement.appendChild(note)
+            } else {
+                noteElement.innerHTML = note || ''
+            }
+            noteElements.push(noteElement)
```
cmdr2, line 286: Do we need the `noteElements` array, given that we only ever push one element into it?

cmdr2: I think it's kept as an array, so than an empty array can be squashed in line 305 (incase `note` is undefined).

How about a common array, into which `labelElement` and `noteElement` (if available) is pushed. And then just use that common array in line 305?

ui/media/js/utils.js
```diff
@@ -708,3 +710,19 @@ function createElement(tagName, attributes, classes, textOrElements) {
...
+/**
+ * Add a listener for arrays
+ * @param {keyof Array} method
+ * @param {(args) => {}} callback
+ */
+Array.prototype.addEventListener = function(method, callback) {
+    const originalFunction = this[method]
+    if (originalFunction) {
+        this[method] = function() {
+            console.log(`Array.${method}()`, arguments)
```
cmdr2, line 723: Do we need this console.log?

cmdr2:
Thanks! Looks good

ui/media/js/utils.js
ogmaresca: @cmdr2 No, I can remove it

ui/media/js/parameters.js
ogmaresca: That would also work

PR #1097, opened 2023-03-31
add the ability to directly merge lora safetensors with models 

ollobrains:
none of the other programs i have come across for stable diffusion have the ability to merge LORA files directly into a model, expanding its abilities with very specific additions.  Both are safe tensors, it would be a great feature if it was easy to do and use.  Currently we get a tensor a doesnt match tensor b size error but a script i did find which is difficult to use.

If something similar that worked could be put into model merging / beta LORA builds it might attract a lot more users, automatic1111 and others do not have it yet.

https://github.com/kohya-ss/sd-scripts  this project does it but i cant get my head around it, a much simplier way must exist.

Thoughts , im sure many more users would get involved, its a much asked for feature on reddit etc, but no one has done it yet


ollobrains:
sorry guys update provide in the 2 issue threads as it stands now

In beta mode (without opt in lora) can run image generation fine
Cannot merge LORA files into main safetensors

In LORA opt in mode - Can merge LORA files into main safetensors 
- Image generation results in a error message

So to merge LORA files into my main model i need to log into lora mode, do the model mixing, then opt out of lora mode back to main beta branch and the image generation works

Hope this works plus the error logs provided in sort it out so that lora mode enables merging and running the image generation.

JeLuF:
Please stop using Pull Requests for ideas and bug reports. File those as **Issue**

PR #1106, opened 2023-04-01
Undo UI cleanup

patriceac:
Currently the entire toolbar is shown when there are no tasks left. This PR just keeps the Undo button visible in that scenario for a more consistent UI. Entire toolbar remains visible when there are still visible tasks.

**Before**
image

**After**
Just the undo button is displayed if there are no tasks left:
image

Entire toolbar remains visible when there are still visible tasks:
image


cmdr2:
Thanks!

PR #1107, opened 2023-04-01
Add LoRA Strength to metadata files

ogmaresca:
For some reason it's showing up between `Sampler` and `Stable DIffusion Model` instead of `LoRA model`.


cmdr2:
Thanks, good catch!

PR #1108, opened 2023-04-02
Increase the random seed range

ogmaresca:
Increase random seed range from 10,000,000 to 4,294,967,295 like Automatic1111.


cmdr2:
Thanks!

PR #1109, opened 2023-04-02
Reset the LoRA dropdown if not present in the task

patriceac:
It's all in the title. This properly reset the LoRA selection to None if no lora is provided in the task.


cmdr2:
Thanks!

PR #1112, opened 2023-04-02
Persistentish user configuration

DianaNites:
This adds support for a `user_config.(sh|bat)` script thats called after confg.sh, to be used to set custom environment options

For example, this script can persistently override `update_branch`, or locally set environment variables at runtime for advanced AMD GPU usage, such as `HSA_OVERRIDE_GFX_VERSION` or `ROCR_VISIBLE_DEVICES`, or to set stuff like `XDG_CACHE_HOME` so that all the large model downloads are persisted.

Unlike the existing `config` scripts, these wont be overwritten

### DIFF
--- a/scripts/on_env_start.bat
+++ b/scripts/on_env_start.bat
@@ -8,6 +8,10 @@ if exist "scripts\config.bat" (
     @call scripts\config.bat
 )
 
+if exist "scripts\user_config.bat" (
+    @call scripts\user_config.bat
+)
+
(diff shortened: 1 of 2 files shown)


cmdr2:
Thanks! :)

A request for future PRs - please open PRs against the `beta` branch. Merging to `main` releases the code instantly to all the users.

Ideally we should've made a separate `release` branch, and used `main` for development. But I was young and foolish :)

PR #1117, opened 2023-04-03
Add back/forward buttons to switch between images in tasks

ogmaresca:
Adds back/forward buttons in the image modal, if there is a previous or next image. The arrow keys also function as the arrow buttons.


cmdr2:
Nicely done, thanks!

PR #1129, opened 2023-04-07
fix font definition for some users

JeLuF:
I don't know why it breaks only for a few users, but this patch seems to fix the issue. https://discord.com/channels/1014774730907209781/1081867905907572746 https://discord.com/channels/1014774730907209781/1093215237307637901/1093223612615491604


cmdr2:
Thanks!

PR #1135, opened 2023-04-07
Don't paste empty prompts

JeLuF:
When pasting e.g. an image, window.clipboardData).getData('text') returns an empty string, which would delete the prompt. 

https://discord.com/channels/1014774730907209781/1093186485563424838


cmdr2:
Thanks!

PR #1138, opened 2023-04-09
Properly reset LoRA selection from Use Settings

patriceac:
Repro steps: create a task with LoRA dropdown set to None, and a second task with a LoRA selected. Then click on Use Settings from the first task.

Expected result: LoRA set to None.
Actual result: LoRA remains unchanged (as selected from second task).


cmdr2:
Thanks!

PR #1142, opened 2023-04-09
Fix the tooltip display for long modifiers

patriceac:
Fixes this: https://discord.com/channels/1014774730907209781/1052771989426814986/1094641806194528266

After checking, this impacts every user, not just people using my plugin, so I think the fix should go in core for everyone to benefit from it.


patriceac:
Posted to Main by mistake.

PR #1144, opened 2023-04-10
Fix the styling of disabled image modifiers

patriceac:
Long custom modifiers in a disabled state (e.g. right-click on the image tag) are properly restored as disabled but are incorrectly shown as "active" in the UI. I somehow missed that in https://github.com/cmdr2/stable-diffusion-ui/pull/1062, so here is the fix for that. This change only applies to plugins that try to restore image modifiers, no change to the regular UI.


cmdr2:
Thanks!

PR #1155, opened 2023-04-13
Don't scan safetensors files in load_default_models()

JeLuF:
`is_malicious_file()` already skips the scanning of safetensors files.
This PR adds a similar check to `load_default_models()`, so that no messages like 

```
ERROR: parsing pickle in [...] at position 0, opcode b'\xcd' unknown
```

are logged any more.

https://discord.com/channels/1014774730907209781/1095087265916657695/1095236389290917898


cmdr2:
Thanks!

PR #1166, opened 2023-04-16
Don't save model_path if initial load fails

### LINKED ISSUE #882: The model for Stable Diffusion has not been loaded yet!
**Describe the bug**
The model has not been loaded. I just succeed so install the UI and I got this message at my first try. Each time I try to generate images, I got the same error.

**To Reproduce**
1. Launch 'C:\stable-diffusion-ui\Start Stable Diffusion UI.cmd'
2. Click on 'Make image'

**Expected behavior**
Generate some beautiful images :)

**Screenshots**
Error - Stable Diffusion

**Desktop:**
 - OS: Windows 10
 - Browser: Google Chrome
 - Version: Version 109.0.5414.120 (Build officiel) (64 bits)

JeLuF:
Fixes #882

If the load of the model fails during the initialization, an attempt to render an image using the same model fails because ED doesn't notice that the model has to be loaded. This PR ensures that the model is being reloaded if the initial load fails. If the second load attempt fails as well, the user will get a more helpful error message than 'model not loaded yet'.

### DIFF
--- a/ui/easydiffusion/model_manager.py
+++ b/ui/easydiffusion/model_manager.py
@@ -49,6 +49,7 @@ def load_default_models(context: Context):
         except Exception as e:
             log.error(f"[red]Error while loading {model_type} model: {context.model_paths[model_type]}[/red]")
             log.exception(e)
+            del context.model_paths[model_type]
 
 def unload_all(context: Context):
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #1171, opened 2023-04-17
Don't download 1.4 if other models are available

JeLuF:
On Linux, the check is recursively. On Windows, it only checks for files in `models\stable-diffusion`, not in the subdirectories.

If the 1.4 model file _does_ exist, it will still check the size.


cmdr2:
Thanks!

PR #1173, opened 2023-04-18
Add "Start scanning..." to getModels()

JeLuF:
Provide a hint to users what ED is currently using.  Use case: User has built an infinite loop using symlinks, ED model scan will never finish.

https://discord.com/channels/1014774730907209781/1097764777553580092


cmdr2:
Thanks!

PR #1182, opened 2023-04-20
Don't write config.bat and config.sh any more

JeLuF:
Introduces scripts/get_config.py.
> usage: get_config.py [-h] [--default=DEFAULT] key [key ...]

If config.yaml exists, it will use that file. If it doesn't exist, it will look for config.json.

Get the update_branch setting from the config:
`python scripts/get_config.py update_branch`

Get net.listen_port:
`python scripts/get_config.py net listen_port`

Get the net.listen_to_network setting, return False if the setting doesn't exist in the config file:
`python scripts/get_config.py --default=False net listen_to_network`

### DIFF
--- a/scripts/get_config.py
+++ b/scripts/get_config.py
@@ -1,0 +1,45 @@
+import os
+import argparse
+
+# The config file is in the same directory as this script
+config_directory = os.path.dirname(__file__)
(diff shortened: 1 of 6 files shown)


cmdr2:
Thanks! I made a slight tweak, to copy the script in on_sd_start (for the first run, when on_env_start will still be the old one)

PR #1191, opened 2023-04-22
Add splash screen for testDiffusers users

JeLuF:
The splash screen will only be shown once.

The splash screen version number can be used to roll out a new splash screen, which will also be shown only once.

Clicking on the EasyAndroidLady icon shows the splash screen again.

image


cmdr2:
Thanks! :)

PR #1203, opened 2023-04-25
Add support for Latent Upscaling

JeLuF:
https://huggingface.co/docs/diffusers/api/pipelines/stable_diffusion/latent_upscale

Depends on https://github.com/easydiffusion/sdkit/pull/32

Option not enabled
image

Option enabled, no extra upscaling prompt
image

Option enabled, with extra upscaling prompt
image

Result image:
image

### DIFF
--- a/ui/easydiffusion/model_manager.py
+++ b/ui/easydiffusion/model_manager.py
@@ -106,6 +106,7 @@ def reload_models_if_necessary(context: Context, task_data: TaskData):
         "gfpgan": task_data.use_face_correction,
         "realesrgan": task_data.use_upscale,
         "nsfw_checker": True if task_data.block_nsfw else None,
+        "latent_upscaler": True if task_data.use_latent_upscaler else None,
         "lora": task_data.use_lora_model,
     }
     models_to_reload = {
@@ -256,6 +257,8 @@ def getModels():
 
     if models_scanned > 0:
         log.info(f"[green]Scanned {models_scanned} models. Nothing infected[/]")
+    else:
+        log.info("[green]No new models found[/]")
 
     # legacy
     custom_weight_path = os.path.join(app.SD_DIR, "custom-model.ckpt")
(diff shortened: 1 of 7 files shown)


cmdr2:
This is great! Thanks :)

Are there any examples of using prompts, seeds, steps (other than the original settings)? I wonder if it'll be possible to keep this as just another entry in the "Upscale" drop-down, with sensible defaults. Like upscale by: "RealESRGAN", "RealESRGAN Anime", "Latent".

I haven't played with this upscaler yet, so it's possible that we really need to provide fine-grained control to the user.

I'm okay with merging this as-is, but if there isn't a lot of demand for fine-grained control, maybe we could start simple and add complexity if the users demand it?

Thanks a bunch for taking this up, nicely done.

JeLuF:
> I'm okay with merging this as-is, but if there isn't a lot of demand for fine-grained control, maybe we could start simple and add complexity if the users demand it?

I was able to change the structure of the skin and fabric by playing around with the prompt. The effects are subtle, and I don't always understand how my prompt impacts the result. 
image
image
See also https://discord.com/channels/1014774730907209781/1014774732018683928/1100402973697380452

I combine the latent upscaler with the realesrgan upscaler to get a higher resolution. Adding it to the dropdown wouldn't allow to do this.

Should the manual settings be hidden behind an "Auto" setting?
image

patriceac:
Why have a separate toggle? Latent upscaling could live in the "Scale up" dropdown, and include a "scale by" dropdown. If scale by > 1, then chain up with ESRGAN upscaler.

JeLuF:
**Convert to draft**

After a longer discussion with Patrice, he and me came to the conclusion that it might make more sense to handle the Latent Upscaler as an img2img model instead of an upscaler filter. The amount of new settings and their similarity to the img2img options drove this decision.

JeLuF:
Closing for now

PR #1224, opened 2023-04-28
Disable "TypedStorage is deprecated" user warnings

JeLuF:
These warnings clog up the logfiles and worry users.

```
G:\stable-diffusion-ui\stable-diffusion\env\lib\site-packages\sdkit\models\model_loader\stable_diffusion\convert_from_ckpt.py:1084: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
  with safe_open(checkpoint_path, framework="pt", device="cpu") as f:
G:\stable-diffusion-ui\stable-diffusion\env\lib\site-packages\torch_utils.py:776: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
  return self.fget.get(instance, owner)()
G:\stable-diffusion-ui\stable-diffusion\env\lib\site-packages\torch\storage.py:899: UserWarning: TypedStorage is deprecated. It will be removed in the future and UntypedStorage will be the only storage class. This should only matter to you if you are using storages directly.  To access UntypedStorage directly, use tensor.untyped_storage() instead of tensor.storage()
  storage = cls(wrap_storage=untyped_storage) 
```

### DIFF
--- a/ui/easydiffusion/app.py
+++ b/ui/easydiffusion/app.py
@@ -5,6 +5,7 @@ import socket
 import sys
 import traceback
 import urllib
+import warnings
 
 from easydiffusion import task_manager
 from easydiffusion.utils import log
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #1228, opened 2023-04-30
Toast notification support

patriceac:
Adding support for toast notifications for use in Core and user plugins.

image

Usage in the code:
showToast('This is a toast notification') // default 5s timeout
showToast('This is another toast notification', 10000); // 10s timeout
showToast('And here is yet another showing an error state', 5000, true); // error state


cmdr2:
Thanks! Very cool :)

PR #1250, opened 2023-05-04
Allow grabbing the image to scroll zoomed in images

ogmaresca:
When the user zooms in on the image modal, it will allow the user to grab the image to drag it around.


cmdr2:
Thanks!

PR #1260, opened 2023-05-09
Unset PYTHONHOME

JeLuF:
If the PYTHONHOME variable is set, python will not use the python libraries installed by ED, but those provided by the user. Unsetting PYTHONHOME ensures that the correct libraries are loaded.

https://discord.com/channels/1014774730907209781/1105238156804108398

### DIFF
--- a/scripts/Start Stable Diffusion UI.cmd
+++ b/scripts/Start Stable Diffusion UI.cmd
@@ -4,6 +4,7 @@ cd /d %~dp0
 echo Install dir: %~dp0
 
 set PATH=C:\Windows\System32;%PATH%
+set PYTHONHOME=
 
 if exist "on_sd_start.bat" (
     echo ================================================================================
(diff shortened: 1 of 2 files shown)


cmdr2:
Just wondering if this will break any installations that incorrectly used PYTHONHOME (and therefore didn't install some packages)..

JeLuF:
That might be the case. We'd need to add a file to the installer (using build.*), that will not created by `git pull`. If this signal file exists, unset PYTHONHOME. Like this, only new installs will disable PYTHONHOME.

We'd need to ship new Linux and Windows installers, which would also solve the PYTHONNOUSERSITE issues some users are experiencing.

https://discord.com/channels/1014774730907209781/1107094235750076536
 https://discord.com/channels/1014774730907209781/1014774732018683928/1107333358251409418

cmdr2:
Thanks!

PR #1284, opened 2023-05-19
Add custom Listening Address to the config file

orion-2481:
Choosing a custom IP Address to host on isn't possible, which doesn't help those who use tailscale and would like to **not** host on the available 2 addresses.

Updated both bash and batch to account for this.

New config file flag (config.json for example)

{"render_devices": "cuda:0", "update_branch": "main", "ui": {"open_browser_on_start": false}, "net": {"listen_port": 30000, "listen_to_network": false, "listen_address": "112.69.241.57"}, "model": {"stable-diffusion": "sd-v1-4"}, "vram_usage_level": "balanced"}

Thanks for the work.


JeLuF:
This PR does not contain any UI to configure these values. Might be OK if we declare this to be an "advanced option", but if the listen_address is set, the "listen_to_network" option is being ignored. Setting the listen_address and disable "listen to the network" in the UI settings will still expose the service on the listen_address.

I'd propose to 
```
if listen_to_network is true:
    if a listen IP is set:
        BIND_IP = Listen IP
    else:
        BIND_IP = 0
else:
    BIND_IP = 127.0.0.1
```
 
I didn't test the PR in its current form, but I have the impression that the Windows version assigns `notaddressed` to `ED_BIND_PORT`, but never changes it to a number. Uvicorn won't like this.

JeLuF:
Implemented with #1405 
Thank you for the suggestion!

=> Closing this PR

PR #1285, opened 2023-05-19
Add Clip Skip to metadata files

ogmaresca:
Also, force the properties to be in a consistent order so that, for example, LoRA strength will always be the line below LoRA model. I've rearranged the properties so that they are saved in the same order that the properties are laid out in the UI


cmdr2:
Thanks!

PR #1286, opened 2023-05-21
Automatically use 'Low' when VRAM<4.5GB

JeLuF:
When starting EasyDiffusion for the first time, initialize vram_usage_level based on the amount of VRAM. If cuda:0 has less than 4.5GB, switch to "low".


cmdr2:
Nice, thanks!

PR #1287, opened 2023-05-21
Network settings - Fix typo.

JeLuF:
The 'Please restart' text should be part of the note, not the label.
### Old
image


cmdr2:
Thanks!

PR #1288, opened 2023-05-22
get_config: return default value if conf file is corrupted

JeLuF:
https://discord.com/channels/1014774730907209781/1110053226709000222

A corrupted config file prevented the application from starting. With this fix, if the file is no valid yaml/json, proceed using the default value.


cmdr2:
Thanks!

PR #1298, opened 2023-05-24
Fix confirmation dialog

JeLuF:
By splitting the confirmation function into two halves, the closure was lost.

https://discord.com/channels/1014774730907209781/1110971052546334791


cmdr2:
Thanks!

PR #1304, opened 2023-05-25
Remove warning when reusing settings - Fixes #1290

### LINKED ISSUE #1290: Remove warning when reusing settings
**Describe the bug**
The warning "Small image sizes can cause bad image quality" does not get removed if you change the settings by "Use these settings" of an earlier image generation.
It only disappears after changing the width or height manually to another value.

**To Reproduce**
1. Enqueue an image (or more) with high enough width and height for the warning not to appear
2. Change the height and width to a setting that causes the warning to appear (for example 256x256)
3. Click on "Use these settings" of the image(s) enqueued in 1
5. See that the warning did not disappear

**Expected behavior**
The warning should disappear.

**Screenshots**

 - OS: Windows, CPU mode
 - Browser: Brave (Chromium)
 - Version: Beta Channel, likely this commit: 29ec34169c13403ebf7780a38e8143f6ea071fe9

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**

JeLuF:
When processing a metadata file, trigger events for changed height/width/prompt fields. Fixes #1290, and also updates the "Make X images" button.

### DIFF
--- a/ui/media/js/dnd.js
+++ b/ui/media/js/dnd.js
@@ -43,6 +43,7 @@ const TASK_MAPPING = {
         name: "Prompt",
         setUI: (prompt) => {
             promptField.value = prompt
+            promptField.dispatchEvent(new Event("input"))
         },
         readUI: () => promptField.value,
         parse: (val) => val,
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #1309, opened 2023-05-26
Add tiling and latent upscaler steps to metadata

ogmaresca:
Also fix txt metadata labels when also embedding metadata


cmdr2:
Thanks!

PR #1313, opened 2023-05-27
Share ED via Cloudflare's ArgoTunnel

JeLuF:
Shares the Easy Diffusion instance via https://try.cloudflare.com/

Since the list of settings has become quite long, I've put the sharing related settings into their own section.

image

image


cmdr2:
Very cool, thanks! :)

PR #1316, opened 2023-05-28
Fix #1312 - invert model A and B ratio in merge

### LINKED ISSUE #1312: The "merge models" feature inverts A with B ratio, compared to what the UI claims
**Describe the bug**
The "merge models" feature uses opposite ratio compared to what UI says and what the user would expect.

**To Reproduce**
Let's take two models:
image
A: Hyper-realistic model
B: Art model, painting style

I then produce two merges:

1: Merge ratio 93%. I expect images to be almost realistic, as it should take 93% from model A. I call this new model "realistic"
image

2: Merge ratio 7%. I expect images to be almost paintings, as it should take 93% from model B. I call this new model "artsy"
image

**Expected behavior**
I expect merged model 1 "realistic" to be almost realistic, and merged model 2 "artsy" to be almost a painting.
What happens instead is the opposite, which makes me thing that the UI or the tooling code at some points inverts the ratio compared to what the UI claims.

image

image

 - OS: Windows 11
 - Browser: Chrome

JeLuF:
From #1312:
> The "merge models" feature uses opposite ratio compared to what UI says and what the user would expect.

### DIFF
--- a/ui/plugins/ui/merge.plugin.js
+++ b/ui/plugins/ui/merge.plugin.js
@@ -403,16 +403,19 @@
                 // Batch main loop
                 for (let i = 0; i < iterations; i++) {
                     let alpha = (start + i * step) / 100
-                    switch (document.querySelector("#merge-interpolation").value) {
-                        case "SmoothStep":
-                            alpha = smoothstep(alpha)
-                            break
-                        case "SmootherStep":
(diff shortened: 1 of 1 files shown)


cmdr2:
Thanks!

PR #1317, opened 2023-05-28
Fix SyntaxWarning on startup

ogmaresca:
Fixes
```
/ssd2/easydiffusion/ui/easydiffusion/utils/save_utils.py:222: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if task_data.use_upscale is not "latent_upscaler" and "latent_upscaler_steps" in metadata:
  ```


cmdr2:
Thanks!

PR #1328, opened 2023-06-03
Allow LoRA strengths between -2 and 2

ogmaresca:
Some LoRAs like https://civitai.com/models/58390/detail-tweaker-lora-lora and https://civitai.com/models/62700/ivorygoldai can get good results with strengths outside of the [0, 1] range


cmdr2:
Thanks!

Issue #1339, opened 2023-06-10
[update] start.sh still referring to https://github.com/cmdr2/stable-diffusion-ui


arrowgent:
**Is your feature request related to a problem? Please describe.**
github handles this natively but your scripts or backend to git pull and update should reflect the new URL
currently it is stated that `https://github.com/cmdr2/stable-diffusion-ui` is the source of the updated files

**Describe the solution you'd like**
change these to `https://github.com/easydiffusion/easydiffusion` since github automatically redirects here

**Describe alternatives you've considered**
editing the scripts myself (which will get overwritten next time i launch the script)

**Additional context**

```
asy Diffusion

Easy Diffusion's git repository was already installed. Updating from beta..
HEAD is now at 16fcb4ed Merge pull request #1314 from JeLuF/dndgan
Already on 'beta'
Your branch is up to date with 'origin/beta'.
remote: Enumerating objects: 202, done.
remote: Counting objects: 100% (202/202), done.
remote: Compressing objects: 100% (85/85), done.
remote: Total 202 (delta 134), reused 177 (delta 114), pack-reused 0
Receiving objects: 100% (202/202), 123.10 KiB | 3.08 MiB/s, done.
Resolving deltas: 100% (134/134), completed with 26 local objects.
From https://github.com/cmdr2/stable-diffusion-ui
   16fcb4ed..924fee39  beta       -> origin/beta
   dd95df8f..7417c1af  cf         -> origin/cf
   3045f521..a2856b2b  main       -> origin/main
Updating 16fcb4ed..924fee39
```

arrowgent:
further infact most of the wiki and internal links are pointing to the previous github pages
`https://github.com/cmdr2/stable-diffusion-ui`

cmdr2:
Hi @arrowgent Yes, the repo was moved just a few days ago, and we need to update all the references. Keeping this issue open until that's done.

Please feel free to submit a PR, we'd really appreciate it! I'm personally not going to be able to take this up immediately, since I'm currently finishing up the new LoRA implementation. Thanks!

cmdr2:
The wiki is editable by the public

PR #1342, opened 2023-06-12
Prevent UI freeze caused by intensive prompts

### LINKED ISSUE #1324: UI Freeze Upon Pasting Prompt
**Describe the bug**
Easy-Diffusion UI Freezes when this text is pasted into the prompt field:
"hyperrealistic full length portrait of gorgeous goddess | standing in field full of flowers | detailed gorgeous face!! | full body!! | skimpy armor | god rays | intricate | elegant | realistic | hyperrealistic | cinematic | character design | concept art | highly detailed | illustration | digital art | digital painting | depth of field"

Pasting part of the prompt and manually typing the rest also results in a freeze. I suspect the "|" has something to do with it.

**To Reproduce**
Paste the above prompt into the prompt field

**Expected behavior**
No immediate effects

 - OS: KDE Neon (5.19.0-43-generic)
 - Browser: Librewolf
 - Version: Firefox 113.0.2-1

**Recording**
https://github.com/cmdr2/stable-diffusion-ui/assets/48196283/a0e7cc34-a018-4d1d-98c5-2255767cdaed

ManInDark:
Re-implemented multiple functions in a way that mimics their original function but only returns how many images would be generated from given prompts. The functions are all named after their original with -Number added at the end.

This fixes #1324 

Although I tried to re-implement all necessary function, but some (like applySetOperator) need to run for it to work.

While doing that, I added some comments, but feel free to remove them as they're specific to this fix, and I left them in to sort of show how / what I did.


ManInDark:
PS: I compared the performance and with this pattern: `1|2|3|4|5|...|n` before it would start to take quite long around n=20 and with this fix it doesn't even sweat at n=30.

PPS: this does not fix the lag if you actually went and sent all the prompts to generate

ManInDark:
Changed all links to now reference easydiffusion/easydiffusion, which closes #1339

ui/media/js/main.js
```diff
@@ -1546,15 +1617,21 @@ heightField.addEventListener("change", onDimensionChange)
...
-        Math.max(parseInt(numOutputsTotalField.value), parseInt(numOutputsParallelField.value)) * getPrompts().length
+        Math.max(parseInt(numOutputsTotalField.value), parseInt(numOutputsParallelField.value)) * getPromptsNumber()
     let imageLabel = "Image"
     if (totalImages > 1) {
         imageLabel = totalImages + " Images"
     }
     if (SD.activeTasks.size == 0) {
-        makeImageBtn.innerText = "Make " + imageLabel
+        if (totalImages >= 10000)
+            makeImageBtn.innerText = "Make tens of thousands of images"
```
cmdr2, line 1627: How about "Make 10000+ images", to prevent the button from becoming too long (especially on mobile, which many users use)? Thanks

ui/media/js/main.js
```diff
@@ -1378,6 +1445,10 @@ function permutePrompts(promptBase, promptMatrix) {
     return prompts
 }
 
+function permutePromptsNumber(promptMatrix) { // this should calculate how many different prompts can be made with the prompt matrix
```
cmdr2, line 1448: Is this function necessary? How about just calling `permuteNumber(promptMatrix)` in line 1419 directly?

cmdr2:
It's okay if some combinations didn't work in the past. For e.g. I don't know if "matrixes in sets" even work today. In that case, it's okay to ignore that scenario.

Just ensuring that we don't have any regressions (runtime errors, incorrect numbers etc). If a scenario didn't work in the past, then we can ignore that for now. Thanks!

ui/media/js/main.js
ManInDark: Wanted to keep the structure sort of the same to make it easier to follow, but applied suggestion in latest commit.

ManInDark:
Tests:
`1|2|3|4 -> 8`
`1, 2, 3, 4 | {5,6} | 7 -> 8`
`{{1,2},{3,4}} -> 4`

All returned the same length with the new function as with the old one.

Not working was `{1|2|3,4}, 5`, which returned 5 instead of 9, but that didn't appear to work correctly as you mentioned (following what getPrompts() returned when called in the console): `['{1', '1, 5', '3, 5', '4, 5', '{1, 2', '1, 5', '2, 5', '3, 5', '4, 5']`

I applied both of your suggestions in the last commit.

ManInDark:
Just for reference, if someone wants to change the 10k+ display later, I don't think you'd have to keep that feature. This is because the pure number calculation is quite quick and only takes longer the more sets there are, but the time it takes just grows linearly instead of exponentially.

cmdr2:
Thanks @ManInDark ! Sorry for the delay, I was away for a few days. Will review this and merge soon!

cmdr2:
Thanks, and sorry for the long delay! Very nice work :)

ManInDark:
If the testing prompts are used again later:
`{1}` should be tested too, as it caused a problem this time

PR #1355, opened 2023-06-19
Image Modifier UI Overhaul

Hakorr:
# Image Modifier UI Overhaul

Fixes and changes to the Image Modifiers and its container. Better flow and more space for clarity and possible future updates regarding modifiers.

### Image Modifier "Dropdown" Button
Relocated image modifiers closer to the prompt for better flow.

image

### Image Modifier "Dropdown" Opened
This is how the image modifier container looks like now.

image
image

### Selected Cards

image

### Settings Popup
Image style and thumbnail size settings can be found alongside custom modifiers.

image

### Miscellaneous

- Additionally fixed the document body overflowing when on mobile.

- The Image Modifier UI overhaul shouldn't affect plugins. The plugins that come with Stable Diffusion by default were tested and should work fine without any updates. The overhaul also works with themes, however the light theme can look a bit ugly, didn't have time to fix that, sorry!

- This overhaul focuses on the UI. The JS side of things still need an overhaul, the current system of locating the cards and such is a bit messy, though has worked for almost a year, so it's not an immediate issue.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -63,8 +63,38 @@
                         <a href="https://github.com/cmdr2/stable-diffusion-ui/wiki/Writing-prompts#negative-prompts" target="_blank"><i class="fa-solid fa-circle-question help-btn"><span class="simple-tooltip top">Click to learn more about Negative Prompts</span></i></a>
                         <small>(optional)</small>
                     </label>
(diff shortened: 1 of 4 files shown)


cmdr2:
Thanks! This is awesome work :)

PR #1362, opened 2023-06-22
Basic embeddings support

JeLuF:
Loads all embeddings from the embeddings folder when loading a mode.
Requires https://github.com/easydiffusion/sdkit/pull/42


ui/media/js/main.js
```diff
@@ -2131,6 +2137,79 @@ document.getElementById("toggle-cloudflare-tunnel").addEventListener("click", as
...
+            if (e.shiftKey) {
+                if (!negativePromptField.value.endsWith(" ")) {
+                    pad = " "
+                }
+                negativePromptField += pad + text
+            } else {
+                if (!promptField.value.endsWith(" ")) {
+                    pad = " "
+                }
+                promptField += pad + text
```
rbertus2000, line 2184:
```suggestion
                if (!negativePromptField.value.endsWith(" ")) {
                    pad = " "
                }
                negativePromptField.value += pad + text
            } else {
                if (!promptField.value.endsWith(" ")) {
                    pad = " "
                }
                promptField.value += pad + text
```
I tested this code on my machine and in order to work properly you need to reference .value to add the token at the end of the prompt or negativeprompt.

cmdr2:
Thanks a lot! And sorry for the delay and back-and-forth on this PR.

PR #1363, opened 2023-06-22
Show COMSPEC variable in logs

JeLuF:
Non-standard COMSPEC settings can cause very odd behaviour on Windows, e.g. permission errors when trying to start `.cmd` files from the Explorer, or issues with nested commands, e.g. in `FOR` loops.

To simplify spotting these issues, I'd like to have the COMSPEC variable displayed in the log output.

The default value is `C:\Windows\System32\cmd.exe`

### DIFF
--- a/scripts/Developer Console.cmd
+++ b/scripts/Developer Console.cmd
@@ -22,6 +22,8 @@ call where conda
 call conda --version
 
 echo.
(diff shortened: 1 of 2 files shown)


cmdr2:
Thanks!

PR #1368, opened 2023-06-26
Modifier settings dialog improvements

JeLuF:
- Use `<dialog>` for the Modifier settings (solves z-order issues)
- Harmonize dialog design. Use similar header style for image modifiers, modifier settings and tiled image download dialogs

image


cmdr2:
Thanks! This is really cool, I learnt something new.

PR #1370, opened 2023-06-27
change event trigger to mousedown

AssassinJN:
This will fix an issue where the box is accidentally closed when attempting to quickly select text in the textarea.


cmdr2:
Thanks!

PR #1376, opened 2023-06-28
Download dialog redesign, moveable dialogs, code cleanup

JeLuF:
image

Style the download dialog like the modifiers setting

Make dialogs draggable. Click on the dialog's window title bar to move the dialog


cmdr2:
Thanks! Very cool :)

Issue #307, opened 2022-10-08
Document title progress indicator


polypoyo:
**Is your feature request related to a problem? Please describe.**
Currently, there is no document title, so most browsers just show the URL in the window title.

**Describe the solution you'd like**
A simple assignment of `document.title` to "N% - Stable Diffusion UI" while processing, and "Done - Stable Diffusion UI" when all images are done

**Describe alternatives you've considered**
The existing ding sound effect is pretty good, but it requires manually enabling auto-play and reloading. It also doesn't solve the title issue.

**Additional context**

PR #1377, opened 2023-06-29
Implementing #307

ManInDark:
This implements #307 by showing how many images are remaining in the title.
This should remain in the beta for a while because for some reason the number sometimes is not right, showing that a batch is completed even though it is still processing. (But I don't know what exactly is causing the issue)

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -464,7 +464,7 @@
                     <div class="clearfix" style="clear: both;"></div>
                 </div>
                 <div id="supportBanner" class="displayNone">
-                    If you found this project useful and want to help keep it alive, please consider <a href="https://ko-fi.com/easydiffusion" target="_blank">buying me a coffee</a> or <a href="https://www.patreon.com/EasyDiffusion" target="_blank">supporting me on Patreon</a> to help cover the cost of development and maintenance! Or even better, <a href="https://cmdr2.itch.io/easydiffusion" target="_blank">donating the actual cost of this software</a>. Thanks for your support!
+                    If you found this project useful and want to help keep it alive, please consider <a href="https://ko-fi.com/easydiffusion" target="_blank">buying me a coffee</a> to help cover the cost of development and maintenance! Thanks for your support!
                 </div>
             </div>
         </div>


ui/media/js/main.js
```diff
@@ -954,6 +959,16 @@ function onTaskCompleted(task, reqBody, instance, outputContainer, stepUpdate) {
     if (isSoundEnabled()) {
         playSound()
     }
+    updateTitle()
+}
+
+function updateTitle() {
+    let img_remaining = [...document.querySelectorAll("div .imageTaskContainer").entries()].map(c => htmlTaskMap.get(c[1])).filter(task => task.isProcessing).map(task => task.numOutputsTotal - Math.max(0, (task.batchesDone - 1) * task.reqBody.num_outputs)).reduce((total, value) => total + value, 0);
```
cmdr2, line 966: Can we break this into multiple lines please? So that it's clearer what each step does, for future developers to grasp the logic at a quick glance? :)

ui/media/js/main.js
```diff
@@ -849,6 +851,9 @@ function abortTask(task) {
             console.error(e)
         }
     })
+    if (task.batchesDone > 0) {
+        document.title = "Stopped - Easy Diffusion"
```
cmdr2, line 855: abortTask will be called by a user, so we don't need to say Stopped in the title bar (the user will already be in the ED tab). I think we can set the title back to "Easy Diffusion".

cmdr2:
Thanks for taking this up! Left two minor comments, looks fine otherwise.

> This should remain in the beta for a while, because for some reason the number sometimes is not right, showing that a batch is completed even though it is still processing

Can you please investigate this a bit more, before this gets merged to beta? Thanks!

ui/media/js/main.js
ManInDark: I've changed it to multiple lines with names that should explain what each line is doing in [ec7553f](https://github.com/easydiffusion/easydiffusion/pull/1377/commits/ec7553fe526bb488485067226b4283499e6eef86)

ui/media/js/main.js
ManInDark: Applied the change in [ec7553f](https://github.com/easydiffusion/easydiffusion/pull/1377/commits/ec7553fe526bb488485067226b4283499e6eef86), but should I change "Completed - Easy Diffusion" to just "Easy Diffusion" too then? Since it would be clear that if there are no more images that it is finished.

ManInDark:
> Can you please investigate this a bit more, before this gets merged to beta?

I've tried to do that, but I have been unable to find where exactly the problem has its root since it's appeared just sometimes and other times not, which makes it a bit hard to diagnose. I think it wouldn't be an issue to merge it into the beta, since someone might find a pattern which could help solve it.

ManInDark:
And I'm not even sure if there is any problem with the code in this PR since the batchesDone is sometimes higher than what has been generated (e.g. it says 2 even though nothing has been generated yet? (according to cmd log)), which suggests that there might be some sort of underlying issue that this is just a symptom of.

madrang:
@ManInDark probably because task.batchesDone only refers to how many task objects got created. You probably still have task.instances that are running/queued on the server.

ManInDark:
@madrang thanks for the info, I've changed the implementation. That should fix the problem now, so it shouldn't be a problem to merge it now @cmdr2

cmdr2:
Thanks @ManInDark and @madrang , I'll take a look soon

PR #1382, opened 2023-06-30
Don't show copy button without clipboard access

JeLuF:
Don't show the "Copy" button for the cloudflare hostname if there's no clipboard access (e.g. when accessing ED via LAN)

https://discord.com/channels/1014774730907209781/1122893199304822858


cmdr2:
Thanks!

PR #1396, opened 2023-07-09
Fix dialog closing when interacting with <select> child

JeLuF:
https://discord.com/channels/1014774730907209781/1021695193499582494/1127653577360425042 

'The.QCC: Embedding window. Switching between at end and at cursor closes the Embedding window'


cmdr2:
Thanks!

PR #1397, opened 2023-07-09
Add Restart needed to diffusers-only fields if you enable diffusers then refresh without restarting EasyDiffusion

ogmaresca:
Also fix LoRA Alpha container being displayed when diffusers is disabled if you had a LoRA model already set

image

I think this would prevent issues like https://discord.com/channels/1014774730907209781/1014774732018683928/1127049610993606717


cmdr2:
Thanks! Clever idea

PR #1398, opened 2023-07-10
Add embeddings to metadata

ogmaresca:
The values will be comma separated

PR #1402, opened 2023-07-11
Fix handling of embeddings with space in their name

JeLuF:
Use the filename to token name function from sdkit in getModels.
Depends on https://github.com/easydiffusion/sdkit/pull/46


JeLuF:
Needs to incorporate changes for "Add embeddings to metadata #1398"

cmdr2:
Thanks!

PR #1403, opened 2023-07-11
Warn about low vram profile when using embeddings

JeLuF:
Warning until https://github.com/huggingface/diffusers/issues/3922 is fixed.

image


cmdr2:
Thanks!

PR #1405, opened 2023-07-11
Add Bind IP option

JeLuF:
Reimplementation of #1284

```yaml
net:
  listen_port: 9000
  listen_to_network: false
  bind_ip: 1.2.3.4
```
Defines an IP to bind to instead of 0.0.0.0
`listen_to_network` must be set to true for bind_ip to have an effect.

### DIFF
--- a/scripts/on_sd_start.sh
+++ b/scripts/on_sd_start.sh
@@ -72,7 +72,7 @@ export SD_UI_PATH=`pwd`/ui
 export ED_BIND_PORT="$( python scripts/get_config.py --default=9000 net listen_port )"
 case "$( python scripts/get_config.py --default=False net listen_to_network )" in
     "True")
-        export ED_BIND_IP=$( python scripts/get_config.py --default=0.0.0.0) net bind_ip)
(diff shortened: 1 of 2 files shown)


scripts/on_sd_start.sh
```diff
@@ -72,7 +72,7 @@ export SD_UI_PATH=`pwd`/ui
 export ED_BIND_PORT="$( python scripts/get_config.py --default=9000 net listen_port )"
 case "$( python scripts/get_config.py --default=False net listen_to_network )" in
     "True")
-        export ED_BIND_IP=0.0.0.0
+        export ED_BIND_IP=$( python scripts/get_config.py --default=0.0.0.0) net bind_ip)
```
cmdr2, line 75: `--default=0.0.0.0) net bind_ip)`

Is there an extra parantheses-end after 0.0.0.0?

JeLuF: Fixed

cmdr2:
Thanks!

PR #1414, opened 2023-07-15
Format LoRA model/strength as comma separated string in text metadata

ogmaresca:
Change text metadata from

```yaml
LoRA model: ['/easydiffusion/models/lora/first.safetensors', '/easydiffusion/models/lora/second.safetensors', '/easydiffusion/models/lora/third.safetensors']
LoRA Strength: [1.05, 1.0, 1.0]
```

to

```yaml
LoRA model: /easydiffusion/models/lora/first.safetensors, /easydiffusion/models/lora/second.safetensors, /easydiffusion/models/lora/third.safetensors
LoRA Strength: 1.05, 1.0, 1.0
```


cmdr2:
Thanks! We will need additional changes to make this work with drag-and-drop right? I think text metadata isn't working with multi-lora anyway right now.

ogmaresca:
> Thanks! We will need additional changes to make this work with drag-and-drop right? I think text metadata isn't working with multi-lora anyway right now.

Yes, that does need to be fixed regardless

PR #1420, opened 2023-07-17
Help texts for common errors

JeLuF:
- Using an Embedding on 'Low'
- Using LORAs and models with mismatching versions


cmdr2:
Thanks!

PR #1423, opened 2023-07-18
Server side object store

JeLuF:
This PR provides a server side object store ("bucket") that can be used to store items in a (hopefully) secure manner on the server side.

On the Javascript side, two functions are provided: 
- `saveObjectToBucket` - Store any kind of object (JSON, text, image) on the server
- `loadObjectFromBucket` - Load a JSON object from the server.

Images stored to the bucket can be directly addressed: ``

Images need to be stored in their binary form, not using the `data:` URL.

This PR is a prerequisite for
- Maintain a list of tokens for the different LORAs (to be displayed in the embeddings tab)
- Thumbnail management for embeddings


JeLuF:
To be replaced by a SQLite based version

PR #1428, opened 2023-07-22
Fix SyntaxWarning on startup

ogmaresca:
Fix

```sh
/ssd2/easydiffusion/ui/easydiffusion/utils/save_utils.py:225: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif key is "use_embedding_models" and using_diffusers:
```

logs on startup


cmdr2:
Thanks!

PR #1430, opened 2023-07-22
Display Test Diffusers checkbox when enabling beta

JeLuF:
Remove the need to reload the page after enabling beta. Make diffusers section directly available after enabling beta.


cmdr2:
Thanks!

PR #1439, opened 2023-07-28
fix upscale button

rbertus2000:
fixes KeyError with upscale button:
image


cmdr2:
Thanks! I've actually replaced the Upscale button implementation with a different one, so that'll also fix this error. Plus I've fixed the error in the python conversion code. So unfortunately this PR is no longer valid.

PR #1441, opened 2023-07-29
Support more image sizes

JeLuF:
With diffusers, width and height must be a multiple of 8 (instead of 64), allowing more resolution values. This PR replaces the `<select>` boxes with `<input>` boxes. If diffusers are allowed, the step size is 8, otherwise 64.
image

It also adds a popup with buttons to increase the width/height by factors of 1.5, 2 and 3. This popup also provides a list with the 8 most recently used image dimensions.


JeLuF:
Added a swap width/height button
image

cmdr2:
Thanks! Really useful :)

fdwr:
The most recent list looks quite convenient, as I often switched back and forth between two different resolutions.

PR #1476, opened 2023-08-05
Support including Metadata in images shown on UI

### LINKED ISSUE #467: Save image parameters in metadata
Maintaining the parameter information together with an image is crucial for image characterisation and regeneration.

So far, image parameters are partly stored in the filename which is limited by filename length. Using metadata encoding like IPTC or EXIF would allow to store not only selected but all parameters together with each file for a full characterisation of the image. Data can be read by many common image software tools.

mdiller:
This adds support for including metadata in images that are displayed on the ui. This means if the user individually downloads a single image, it will have the metadata attached. (Previously the only way to do this was to turn on the 'download all images generated' setting and grab them from that location)

Note that for this pull request I just hooked it up to the same setting because I'm not sure how ya'll would prefer to have it set up, but I figured this is a reasonable first implementation for it. 

Note that for this to work, it needs to have a change in sdkit, which is in this pull request:
https://github.com/easydiffusion/sdkit/pull/54

fixes #467


morovinger:
Upvoted this, highly needed feature

PR #1481, opened 2023-08-06
Adding Gallery feature

ManInDark:
This has to wait until [this branch](https://github.com/JeLuF/stable-diffusion-ui/tree/bucketlite) by @JeLuF is merged, since it uses the sqlalchemy setup that was added there (I yanked the files for testing in [this commit](92009cff2400255a7b5e0d47553c8fbafe621b3c)).

At the moment, all generated images are automatically added to a database that holds all information about them (Path, Seed, Prompt, Negative Prompt, …). From there they can then be viewed in the newly added Gallery tab, where all images are shown at the moment (could be improved to allow searching / filtering later).

Right now, there is no way to delete entries in the database from the UI, but that might be added later (if access control were to be added to Cloudflare tunnels, you wouldn't want others to delete the images).

Security was considered in the image serving function to only show images that are in the database, you wouldn't want to make it possible to view _every_ image on the device EasyDiffusion runs on.


ManInDark:
Note: for it to work, the image files actually need to be saved, if no files are saved it won't work

ManInDark:
PR for the required branch is #1483

PR #1486, opened 2023-08-08
Change Cursor type on logo hover

ManInDark:
Change cursor type on logo hover to cursor used with buttons to indicate that it can be clicked.


cmdr2:
Thanks!

ManInDark:
You're welcome! 👍🏻

PR #1490, opened 2023-08-11
Resize slider for the advanced image popup

JeLuF:
image

Proportionally resize the image width/height using a slider!


cmdr2:
Thanks! Very cool :)

PR #1491, opened 2023-08-11
Pythonize the uvicorn startup

JeLuF:
- Move half of on_sd_start.{bat,sh} to python
- Detect AMD GPUs and install the right ROCm version (5.2 or 5.5nightly)
- set up the environment variables for AMD GPUs (HSA override, force fp32)


cmdr2:
Thanks! This is really useful :)

PR #1493, opened 2023-08-11
Cancel/Stop/Remove task buttons

JeLuF:
So far, the button to remove a not yet rendered and a completed task was labeled 'Remove'. This can lead to confusions. This PR changes the label to 'Cancel' for not yet rendered tasks. It also changes the color of the undoable 'Remove' button
image


cmdr2:
Thanks, that makes sense. Okay with the label change.

Re color - while I understand that Remove is undoable, it's still a "red" kind of operation, i.e. a delete or remove kind of operation. The fact that it's undoable is an implementation aspect, built for convenience (vs a confirmation popup). It doesn't change the nature of the button IMO.

cmdr2:
Thanks!

PR #1503, opened 2023-08-16
Warn when no controlnet model is chosen

JeLuF:
- Show a warning when a controlnet image has been chosen, but no model
- Highlight the seed, inference steps or controlnet model dropdown if their validation fails.
image


cmdr2:
Thanks!

PR #1504, opened 2023-08-16
Error messages for SDXL embeddings and SDXL yaml files

JeLuF:
image


cmdr2:
Thanks! Edited to remove the SDXL embeddings message (since we now support embeddings with SDXL)

PR #1509, opened 2023-08-18
Improvements/Fixes for embeddings UI

JeLuF:
- Don't show "Use as thumbnail" if no embeddings were used in the prompt
- Fix layout issue on small screens
- use req.use_embeddings_model instead of parsing the prompt
- Add button to upload a thumb" ../../index.html ../css/main.css main.js


cmdr2:
Thanks!

PR #1510, opened 2023-08-18
API to get SHA256 of a model file

JeLuF:
To be used from javascript to collect metadata from civitai 
https://civitai.com/api/v1/model-versions/by-hash/0A35347528

### DIFF
--- a/ui/easydiffusion/server.py
+++ b/ui/easydiffusion/server.py
@@ -139,6 +139,10 @@ def init():
     def modify_package(package_name: str, req: dict):
         return modify_package_internal(package_name, req)
 
+    @server_api.get("/sha256/{obj_path:path}")
+    def get_sha256(obj_path: str):
+        return get_sha256_internal(obj_path)
+
     @server_api.get("/")
     def read_root():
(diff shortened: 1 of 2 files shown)


ui/easydiffusion/server.py
```diff
@@ -451,3 +455,26 @@ def modify_package_internal(package_name: str, req: dict):
         log.error(str(e))
         log.error(traceback.format_exc())
         return HTTPException(status_code=500, detail=str(e))
+
+def get_sha256_internal(obj_path):
+    import hashlib
+    from easydiffusion.utils import sha256sum
+
+    path = obj_path.split("/")
```
cmdr2, line 463: Do we need to check for backslash as well? I've noticed that the model paths with nested subfolders can contain them.

cmdr2: Or we could ensure that the JS side never sends backslashes

cmdr2:
Thanks! This will be a very handy feature

ui/easydiffusion/server.py
JeLuF: /get/models returns a JSON with nested objects, without any separator. So it's the javascript side that needs to make this into a path.

PR #1514, opened 2023-08-20
fix lora-prompt-parser-plugin

rbertus2000:
A little fix to make lora-pormpt-parser compatible with latest lora changes.


cmdr2:
Thanks @rbertus2000 ! I've actually converted the LoRA models section into a "component", so we can assign it an array of model names and weights, and it'll automatically assign the correct elements (and resize if necessary).

Just pushed a fix for this: https://github.com/easydiffusion/easydiffusion/commit/18049d529a8fa305adf207d6387195d4e93757af

Thanks for the PR!

PR #1516, opened 2023-08-21
fix embedding parser and use standard embedding variable for metadata

rbertus2000:
This PR does two things:
1. fix embeddings parser: The current parser will find `easynegative`. but it will ignore `(easynegative)0.8`
2. use the new use_embeddings_model varable for metadata instead of the custom one


cmdr2:
Thanks! Good catch :)

PR #1525, opened 2023-08-27
Remove safetensors hack

JeLuF:
We should upgrade to 0.3.3, since it has wheels for all our supported plattforms.


cmdr2:
Done, thanks! Also updated sdkit's setup.py, for beta (and upcoming main)

PR #1528, opened 2023-08-29
inputmode=numeric/decimal for <input> fields

JeLuF:
Use [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) hints for mobile browsers.

On my phone, the "decimal" keyboard gets used for all of the input modes, so I'm not sure whether this will fix Putan's issue.

https://discord.com/channels/1014774730907209781/1145441618971873310


cmdr2:
Thanks!

PR #1549, opened 2023-08-31
feature: add a label to the task item html that indicates number of i…

bZichett:
Without this commit, it seems the UI simply does not surface this value until the task begins processing.  I will also say this is one of my favorite aspects of easydiffusion (its queuing system is very well done and integrated.)

image

Questions:
Dev decision made to not show when ✕1, but this offsets the prompt differently from ✕2 or more from ✕1.  Can always include this number (✕1) to 'fix' this minor UX issues, and maybe make it more washed out color wise (harder to see, more like a placeholder unless multi image.)  OTOH, it may be nice to know that a task _is_ going to produce batch images, but perhaps more important is prompt alignment (since many times, the prompt is the same in the beginning)

Note: Also included is a minor tweak.  Not sure if it's a plugin I installed or if others see it differently but also included is a css patch for overriding the `min-width: 50px` from `.drag-handle` conflicting css.  Could also rename the class to something different but for now this works fine.  Without this fix it looks like this:

image

### DIFF
--- a/ui/media/css/main.css
+++ b/ui/media/css/main.css
@@ -715,15 +715,26 @@ div.img-preview img {
 .header-content > .drag-handle {
     min-width: initial !important;
 }
-.taskStatusLabel, .taskNumberOfImagesLabel {
+.taskLabel {
     font-size: 8pt;
     background:var(--background-color2);
     border: 1px solid rgb(61, 62, 66);
     padding: 2pt 4pt;
     border-radius: 2pt;
-    margin-right: 5pt;
     display: inline;
 }
+.taskStatusLabel {
+    margin-right: 5pt;
+}
+.multiNumberOfImages {
+    font-weight: bold;
+}
+.taskNumberOfImagesLabel {
+    min-width: 30px;
+    display: inline-block;
+    text-align: center;
+}
+
 .activeTaskLabel {
     background:rgb(0, 90, 30);
     border: 1px solid rgb(0, 75, 19);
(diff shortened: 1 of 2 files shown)


bZichett:
After thinking about above comments a bit more, probably a better UX to just always render the label (✕1).

bZichett:
Ex: 
image

JeLuF:
Wouldn't it be sufficient to show this info in the taskinfo section, similar to seed, dimension, etc?

bZichett:
Yes, @JeLuF.

> Wouldn't it be sufficient to show this info in the taskinfo section, similar to seed, dimension, etc?

I had thrown a bunch of ideas above, and it really comes down to preference.  I prefer it a bit more visible since I render with a lot of multi image jobs, and I like it looking like the Waiting / Processed label.  That being said, I wouldnt mind it being next to the seed either.  It just needs to be _somewhere_ (imo)

This commit can easily be changed to whatever the best UX compromise is.

q5sys:
+1 on this, this is one of those small improvements I've been hoping would come for a while.

cmdr2:
Sorry, I think this is best done as a plugin. I see the utility, but I haven't heard anyone else ask for it all these years. And every extra bit of information added to the screen increases the visual overload.

I'll definitely keep this in mind, maybe in a future re-design. But for now, it's not broadly asked for (not enough) as a need to justify the additional cognitive load for all users. Thanks!

PR #1560, opened 2023-09-01
Error handling for models_dir

JeLuF:
- If the models_dir can not be created (permissions, drive doesn't exist), show an error message providing hints what went wrong:
image

- If the `models_dir` entry in the yaml file is empty, use the default `easydiffusion/models` directory.
    https://discord.com/channels/1014774730907209781/1014774732018683926/1147249736072577104


cmdr2:
Thanks! :)

PR #1563, opened 2023-09-03
AMD/Linux: Warn about file permissions

JeLuF:
- Check whether the user has access to the Kernel Fusion Driver `/dev/kfd`, which is a component of the AMD ROCm stack.
- removed unused environment variable


cmdr2:
Thanks!

PR #1588, opened 2023-09-14
Loramanager fixes

JeLuF:
- avoid console errors in python and JS code
- suppress localhost:9000/null links

### DIFF
--- a/ui/easydiffusion/bucket_manager.py
+++ b/ui/easydiffusion/bucket_manager.py
@@ -55,8 +55,13 @@ def init():
             return bucketfiles
 
         else:
-            bucket_id = crud.get_bucket_by_path(db, path).id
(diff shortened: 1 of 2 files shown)


Menno5:
I'm not really known with Github so I don't know if this can be placed here or that is should be placed in the Issues part?
I place it here now since the subject is about Loramanager fixes.

I don't know if it's something that still has to be implemented into the Lora Manager or if it needs a fix.
So just in case: everything works great but there is no way to add or import thumbnails. The _Upload new thumbnail_ button is clickable but does nothing.
When going to the _Model Tools_ tab, this comes in the log with some errors. See attached log.txt
[log.txt](https://github.com/easydiffusion/easydiffusion/files/12613944/log.txt)

cmdr2:
Thanks!

PR #1604, opened 2023-09-23
FIX: ROCM download URL

flavioislima:
This fixes a typo on the URL, a spacing, that is making the module to fail to install on systems with AMD GPUs


cmdr2:
Thanks for catching this! :)

flavioislima:
> Thanks for catching this! :)

No problem, thanks for maintaining EasyDifusion! 😁

PR #1747, opened 2024-03-14
feat: gallery & document progress indicator

ManInDark:
Includes:
- Gallery
- implementing #307
- Bucket API


ManInDark:
I've done some tests and as far as I can see most of the stuff works, only some newer features like text encoder, scheduler, lora and output_format are not properly restored. Though some may not even be even be accessible in the save option, I'll have to investigate that.

ManInDark:
Since those parameters didn't exist back then, everyone that had already used the gallery feature before needs to execute this SQL in the `bucket/bucket.db` sqlite database:
```SQL
ALTER TABLE images ADD COLUMN use_text_encoder_model VARCHAR DEFAULT '';
ALTER TABLE images ADD COLUMN scheduler_name VARCHAR DEFAULT '';
```
This could be part of an announcement or a migration guide.

ManInDark:
As far as my testing has shown, all attributes except number of images and output format are now saved and restored. Lora does currently not work, since the stored data is in another format than the one required for the frontend. This could potentially be solved later by manually matching the strings to the paths.

ManInDark (commented):
should be fine to merge into beta now, further stuff can also be fixed later

codeCraft-Ritik (commented):
Nice work! The implementation looks clean and easy to follow

PR #1770, opened 2024-04-26
Update lora-prompt-parser.plugin.js

Skoteinos1:
Issue: 
When copy/pasting prompt with sub-prompts and lora, lora is extracted, but sub-prompts are merged into one big prompt.

Because I use script to copy/paste bunch of prompts with different lora's and let it run overnight, to test which lora is the best.

Solution:
To keep /n in prompts. I removed this part:
.replace(/(\s*,\s*(?=\s*,|$))|(^\s*,\s*)|\s+/g, ' ')


Skoteinos1:
feeling happy

OttonelliCarlos (approved):

cmdr2:
Sorry for the extremely late merge. I was on a break. Thanks for the contribution!

PR #1801, opened 2024-06-06
Update torch packages to ROCM 5.7 for Navi 3

tannisroot:
People with RDNA3 (Navi 3) cards have trouble getting Easy Diffusion up and running (see https://discord.com/channels/1014774730907209781/1212916906835845160), and to get it to work it requires a newer ROCM version.
It was reported that while latest 6.0 has some issues with easy diffusion, 5.7 works fine, so this PR updates the ROCM version that gets installed to 5.7 for this generation of AMD cards. The package versions are the latest ones that are available for ROCM 5.7.
I've personally tested this on an RX 7900 XTX and it works (along with another change, PR incoming for that).

### DIFF
--- a/scripts/check_modules.py
+++ b/scripts/check_modules.py
@@ -56,16 +56,16 @@ def install(module_name: str, module_version: str):
         amd_gpus = setup_amd_environment()
         if module_name == "torch":
             if "Navi 3" in amd_gpus:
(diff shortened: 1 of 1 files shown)


tannisroot:
Actually, I'll reopen this with the beforementioned change instead, seeing how it won't work without it, sorry.

PR #1867, opened 2024-12-10
Tjcomserv patch 1 pydantic

tjcomserv:
To fix issue 1866

https://github.com/easydiffusion/easydiffusion/issues/1866

1. fastapi updated to 0.115.6 to be better compatible with pydantic v2.10
2. Update setuptools allowed versions to 69.5.1
3. Add ":bool" to three variables in types.py

I have made this changes locally to test and all seems well.


cmdr2:
Thanks! :)

PR #1874, opened 2024-12-22
Remove inpaint size limit

AvidGameFan:
With the limit removed, inpaint can work with image sizes larger than 768 pixels.

### DIFF
--- a/ui/media/js/image-editor.js
+++ b/ui/media/js/image-editor.js
@@ -559,12 +559,12 @@ class ImageEditor {
         }
 
         if (width > height) {
-            var max_size = Math.min(parseInt(window.innerWidth * 0.9), width, 768)
+            var max_size = Math.min(parseInt(window.innerWidth * 0.9), width)
             var multiplier = max_size / width
             width = (multiplier * width).toFixed()
             height = (multiplier * height).toFixed()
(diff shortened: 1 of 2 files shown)


AvidGameFan:
I discovered a problem with the last commit -- for some inpainted images (perhaps along with "draw"), the width and height are not set to standard resolutions, leading to invalid (blank) resolution fields.

AvidGameFan:
I'm having trouble reproducing my previous problem.  I've been doing inpainting and draw without issues.

AvidGameFan:
After more testing, it appears that when the resolution gets over a threshhold (probably the size of the browser window), it will no longer allow the height/width to be assigned.  Which is odd, since you can just manually key it in.

AvidGameFan:
This final change removes all the limits on the size of the image editor.  This allows the inpainting to work without error, but can result in a very difficult view, with the editor far larger than the screen.  This can be mitigated to a large extent if the user changes the browser zoom to 75% or even 50%.

AvidGameFan:
I've been using these changes, and they are working pretty well.  I am not planning further updates for now.

cmdr2:
Thanks, taking a look! Really appreciate these changes!

PR #1910, opened 2025-03-07
Update on_sd_start.sh

emrul:
Added flag to skip starting uvicorn and waiting for keypress if `INSTALL_ONLY` environment variable is set to `1`. This is to support non-interactive installation (e.g. for use in a Dockerfile)


cmdr2:
I'm closing this PR since it's diverged quite a bit from what was discussed. Please feel free to experiment in your branch, and submit another PR after it's been tested out with the different scenarios. Thanks

PR #1975, opened 2025-09-19
Removed references to Python 3.8 in unix startup files.

jdumke:
on_env_start.sh and developer_console.sh still exported PYTHONPATH for python 3.8, which is outdated, replaced with a more flexible approach. Also altered the shebangs to #!/usr/bin/env bash. 
TODO: fix the same issue in Windows startup files, I can't do ist, I havn't a testing environment for MS.

### DIFF
--- a/scripts/developer_console.sh
+++ b/scripts/developer_console.sh
@@ -1,4 +1,15 @@
-#!/bin/bash
+#!/bin/env bash
+#setup legacy enviroment
+if [ -e "installer_files/env" ]; then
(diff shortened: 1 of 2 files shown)


cmdr2:
Thanks for the PR, and sorry for the extreme delay in looking into this. I've merged it with a few small tweaks, thanks!

cmdr2:
Also fixed a few typos: https://github.com/easydiffusion/easydiffusion/commit/f33af7165a4f7467b108ac733648d3501d8c8458

PR #2021, opened 2026-03-21
fix: add duplicate LoRA detection to prevent garbled images

### LINKED ISSUE #1995: duplicate lora breaks generations
**Describe the bug**
If you select two identical Lora

Image

 images for generation, the generation process breaks. The image becomes a mosaic, a kaleidoscope, or some other incomprehensible jumble.

However, the program does not issue a duplicate warning! You may not notice two identical names; perhaps you placed them in two different folders or accidentally selected the same ones. But this breaks the generation process!

**To Reproduce**
1. Select Lora 1,
2. Select the same Lora 1 again,
3. Click Generate.

**Expected behavior**
we get nonsense from layers of different images.

**Additional context**
Disable the selection of duplicate lora by checking it when creating tasks, or let the program delete them automatically, or let it warn the user about duplicates, and that they might receive nonsense that doesn't match the request!

majiayu000:
Fixes #1995

## Summary

Selecting the same LoRA model twice produces garbled images with no warning. This adds duplicate detection at two points:

1. **Task creation path** (`main.js`): After reading LoRA model selections, deduplicate using a Set — keeps the first occurrence's weight and drops subsequent duplicates. Shows a toast warning when duplicates are removed.

2. **Prompt parser** (`lora-prompt-parser.plugin.js`): Before pushing parsed LoRA tags to the matches array, checks if a LoRA with the same name already exists. Skips duplicates and shows a toast.

## Test Plan

- Add the same LoRA model twice via the UI dropdown — verify toast appears and only one instance is submitted
- Paste a prompt with duplicate LoRA tags like `<lora:model:0.5> <lora:model:0.8>` — verify deduplication and toast
- Verify non-duplicate LoRA combinations still work correctly


cmdr2:
Thanks @majiayu000 ! I've merged the fix (after converting the toasts to console.log, since it didn't feel urgent enough to show a popup and disturb the user).

I've merged it to the `beta` branch, and will release it to the `main` branch after beta users test it for a few days. Thanks for your contribution! :)

PR #2057, opened 2026-08-18
Ignore hidden reference images

AvidGameFan:
When the references field is hidden (usually by switching to a model that does not support them), the reference image array still contains images, and will use them, often causing an error.  The workaround is to go back and select an "edit" model, then remove the reference images, then switch the model back again.

Only a quick test was made by adding a reference image, hiding it, then running an SDXL model.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 # What's new?
 
 ## v4 (preview)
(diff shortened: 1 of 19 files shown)


AvidGameFan:
When accepting this change, be sure to do a squash merge commit, as I messed-up my branch and it took a few commits to get back into sync.

cmdr2:
Thanks! :)

